PR #18, opened 2018-08-04
Rename dungeon locations

andirigible:
Rename dungeon locations referencing map/compass chests & Makar's cell


andirigible:
The sqlite thing was an accident... I didn't mean to commit my Visual Studio stuff but GH Desktop had other ideas, I guess.....

LagoLunatic:
Thanks!

For future reference, you have an old version of item_locations.txt. You should pull from master so you have the latest version before making changes to avoid conflicts.

PR #20, opened 2018-08-06
Animation speedups and grapple turning

TrogWW:


TrogWW:
The only files that should be updated are tweaks.py and randomizer.py. 
Updated animations:
-block pushing (for all block types)
-grapple throwing speed
-grapple wraparound
-rope turning speed increase
-ladder/vine climbing speed increase
-npc focus camera

PR #25, opened 2018-08-10
Support dual (2.x, 3.x) python installs and streamline dependency installing

schwjm:
## What does this pull request change?
* Calling `build.bat` with python 2.x in your default path wants to run pyinstaller with python 2.x, this PR explicitly makes `build.bat` invoke with python 3.
* If `pyinstaller` fails, `build.py` will not run.
* Adds a set of batch files to streamline the installation of dependencies for Windows users. This allows bypassing the command prompt in the happy path if .py files are associated to run with Python 3.

## Testing
- [x] tested both new batch files and error checking from an environment with no packages installed other than pip and setuptools
- [x] tested build.bat changes with empty build environment, verified EXE ran

## Other thoughts
I could update `README.md` to use the batch files, but 2 of the batch files are now explicitly Windows flavored, while the readme currently spells out what you need regardless of platform. Alternatively, I could add .sh variants.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -1,3 +1,3 @@
-
 py -3 -m PyInstaller wwrando.spec
+@if %errorlevel% neq 0 exit /b %errorlevel%
 py -3 build.py
--- a/build.bat
+++ b/build.bat
@@ -1,4 +1,3 @@
 
-pyinstaller wwrando.spec
-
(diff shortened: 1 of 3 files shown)


LagoLunatic:
I updated the commands in build.bat and readme.md to explicitly reference the python version.

PR #114, opened 2019-03-05
added build script for mac

Candygoblen123:
To make it more straightforward for people who use Linux/mac to be able to build

### DIFF
--- a/buildmac
+++ b/buildmac
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+cd "$(dirname "$0")"
+
+
+python3 -m PyInstaller wwrando.spec
+python3 buildmac.py
+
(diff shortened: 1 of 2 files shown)


Candygoblen123:
(it may work with Linux, but i have no way of testing

LagoLunatic:
This script is almost identical to the windows one, only difference is the .exe file extension was removed, so it would be better to modify the existing script to detect the current operating system and only add .exe if it's on windows. This way if the script needs to be modified in the future the same change doesn't need to be made to 2 separate files.

Also just curious, what's the point of running the build script on mac, as opposed to running it from source? Is it to save hard drive space by by deleting the dependencies needed to run it from source?

Issue #106, opened 2019-02-01
Suggestion: Start With Bigger Wallet


Grip555:
A lot of other Zelda randomizers start you off with either the second wallet or let you have 9999 rupees capacity from the start. It's also a pain when you get a seed that means you get either wallet upgrade very late on.

Issue #98, opened 2019-01-21
[Suggestion] Select Gear you Start with in Logic


P3rf3ctXZer0:
One thing I am sort of confused about is why we cannot choose gear we start with? Yes this does fall back to my first request but maybe someplayer would rather start with bombs or the bow. If it is possible can you allow us to choose the gear the seed spawns with?

Grip555:
Yeah, I'd like to start with a better wallet, since it could be a huge pain if you get the wallets late in the game, meaning Triforce Charts and more expensive items need to be got by grinding instead of via playing the game.

PR #115, opened 2019-03-17
Let players choose some additional items to begin the game with

nbouteme:
This add a new tab in the GUI, as it takes quite a lot of space to put it in the main tab, but is a waste to let it have its own tab: some things could be moved or added to this new tab to make it feel less empty.

Default settings

You can select multiple items and move the selection from one column to another using the arrow buttons.
Logic seems to properly handle this.
`custom_funcs.asm` was changed to allocate a fixed size array that can hold up to 32 function pointers, one for each item to add to the inventory, which is currently sufficient but kind of a waste at the same time, I'm not sure what kind of stance you have on this approach, and I'd gladly accept any criticism you have to offer.

This is meant to address issues #106 and #98


LagoLunatic:
Some things I noticed from a quick look.
* If the user selects race mode, swordless mode, and start with 8 triforce shards, usually race mode would put the 3 bows + hookshot as the 4 boss rewards. But with this feature the user can start with even those items, leaving no possible items left as boss rewards for race mode. Not sure how to handle this case.
* The custom starting items aren't factored into the permalink.
* The GUI window was made larger and it now boots up onto the "Starting Gear" tab instead of "Randomizer Settings", I assume both of these were done by mistake.

>Not sure how the Progressive Bomb Bag and Quiver play with their
>respective items themselves so to be safe they aren't supported yet. 

Bomb bag and quiver are unrelated to bombs and bow, they just increase what your carrying capacity will be whenever you get the real item.

>custom_funcs.asm was changed to allocate a fixed size array that can hold up to 32 function pointers, one for each item to add to the inventory, which is currently sufficient but kind of a waste at the same time, I'm not sure what kind of stance you have on this approach, and I'd gladly accept any criticism you have to offer.

128 bytes is pretty small but I do like to be extra careful about how much space is used since we know from custom models that a few hundred kilobytes is all we have to work with before the game starts crashing.

The way I would do it is make each item slot a byte which is the item ID, and have the asm simply call `execItemGet`, which is a vanilla function that handles reading from the list of item get function pointers and calling the appropriate function.
(In this case, 0xFF would mark the end of the list instead of 0 since 0 is a valid item ID.)

nbouteme:
Thank you for your feedback.

> If the user selects race mode, swordless mode, and start with 8 triforce shards, usually race mode would put the 3 bows + hookshot as the 4 boss rewards. But with this feature the user can start with even those items, leaving no possible items left as boss rewards for race mode. Not sure how to handle this case.
Maybe this should be disabled when using race mode. I think players who do race mode don't need/want this activated.

> The custom starting items aren't factored into the permalink.
I didn't do that because a naive approach could make permalinks very long. Do you think there should be a character limit for permalinks?

> The GUI window was made larger and it now boots up onto the "Starting Gear" tab instead of "Randomizer Settings", I assume both of these were done by mistake
Indeed, my bad. I assumed the GUI booted up to this tab just because it remembered the last opened tab.

> `execItemGet`
Oh, I had not seen this. I will rewrite the code to use it, then.

Thanks.

LagoLunatic:
>Maybe this should be disabled when using race mode. I think players who do race mode don't need/want this activated.

You know that probably makes sense.
That sounds good, there will just need to be some kind of indication in the UI that starting gear gets disabled when race mode is checked.

>I didn't do that because a naive approach could make permalinks very long. Do you think there should be a character limit for permalinks?

Yeah, I would like to keep the permalinks on the short side if possible. 32 slots * 8 bits per item ID is gonna be too much.
How about this: Each item index in the inventory.py list gets one bit. So telescope is the first bit, magic armor the second bit, etc, which should keep the number of bits fairly low and only add about 4 chars to the permalink.
Items that appear multiple times like bow would need a bit for each occurrence.

nbouteme:
> some kind of indication in the UI that starting gear gets disabled when race mode is checked.

I'll try to hide it when race mode is checked.

> So telescope is the first bit, magic armor the second bit, etc, which should keep the number of bits fairly low and only add about 4 chars to the permalink.

I'll do that, then.

LagoLunatic:
Fixed some bugs, still a couple others remaining I noticed:
* If the user selects swordless mode and tries to start with hurricane spin there's an error since hurricane spin doesn't exist in swordless
* settings.txt can be manually edited to add whatever items the user wants. This can result in the user starting with more than 31 items, which makes the game crash on boot since some of the asm code got overwritten with item IDs. It can also allow the user to start with items not in the list of options (some of which don't function correctly like rupees). So `randomized_gear` and `starting_gear` will need to be sanitized.

LagoLunatic:
One more:
* Clicking "reset all settings to default" causes not just the starting gear list to empty out, but also the randomized gear list. This makes both lists permanently empty so the user can no longer select any starting items (until they delete/edit their settings.txt manually to fix it)

nbouteme:
Anything else I missed?

LagoLunatic:
I don't think so. Nice job!

nbouteme:
Thanks.

Issue #124, opened 2019-04-06
[Suggestion] More items in Starting Gear


JarheadHME:
I see commented in inventory.py that it's because "we don't want all items to be accessible early and remove too much fun", but I feel like that should be up to the player no? If they want to start with everything and they find it not fun surely they could just remove those items.

The main reason why I'd like to be able to add majority of progress items is so people who want to test things and just want a seed with a full inventory to start with.

I've already made a local version that does this pretty much fine, it's just as simple as adding the specific items to inventory.py, though the only thing I can see that is kinda relevant that doesn't work is the magic meter upgrade. The randomizer doesn't throw an error when generating the iso, but upon making a new save file you still have the small magic meter.

PR #126, opened 2019-04-14
Add more items to optional starting gear

wooferzfg:
I wasn't able to add any of the bags or the Ghost Ship Chart without getting crashes, and the Magic Meter upgrade doesn't actually give you double magic, but the rest of the items seem to work.

For sword upgrades, we filter these out if you have Swordless or Randomized Sword mode on, or if race mode would place it as a dungeon reward.

#124

### DIFF
--- a/wwr_ui/inventory.py
+++ b/wwr_ui/inventory.py
@@ -5,6 +5,7 @@ REGULAR_ITEMS = [
   "Telescope",
   "Magic Armor",
   "Hero's Charm",
+  "Tingle Tuner",
   "Grappling Hook",
   "Power Bracelets",
   "Iron Boots",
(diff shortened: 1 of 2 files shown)


JarheadHME:
Where were the bags and Ghost Ship Chart getting crashes? I added them on a local version about a week ago and they seemed to worked fine...
https://i.imgur.com/2PdVqOw.png
https://i.imgur.com/J3Fx3O5.png
https://i.imgur.com/3ejYAOd.jpg
https://i.imgur.com/JTlMouS.jpg

wooferzfg:
I was getting errors in Dolphin earlier but now I'm not? idk lol, I'll add those items too

LagoLunatic:
The asm for giving starting items only had slots for up to 31 items allocated, and this change pushed the number of total items above that, so if you started with all the items in the list it would overflow the list and start overwriting code with item IDs. Just a coincidence it only happened to woofer with the bags and ghost ship chart.

Magic meter upgrade tries to add to the GUI magic meter, which doesn't exist when starting a new game, so it doesn't work as a starting item. Don't feel like fixing that.

wooferzfg:
@LagoLunatic could you also add Moblin's Letter, Maggie's Letter, Note to Mom, and Cabana Deed to the optional starting gear?

LagoLunatic:
Done.

Also I realized the permalink doesn't work with the empty bottles. Things in the REGULAR_ITEMS list it only remembers if you're starting with 0 of them or all of them, nothing in between. Things in the PROGRESSIVE_ITEMS list can support 0-3 but not 4, so I'm just reducing the number of bottles you can start with to 1 to simplify things.

PR #127, opened 2019-04-14
Allow charts and keys to be used for hints

wooferzfg:
This helps prevent errors when your starting gear includes too many items.


LagoLunatic:
Doesn't seem this pull fixes the error, at least not on permalink `MS42LjAtQkVUQV8yMDE5LTA0LTA3Xzg5NGI4Y2MAVGVzdGluZwBvAxUAR1Ag/3//3xUB`

wooferzfg:
It doesn't work for that permalink because the only progress item left is an Empty Bottle. A potential solution would be to use the same hint for all the fishmen and for the Great Fairy if there's only 1 progress item available to give hints for?

PR #129, opened 2019-04-19
Simplify the Battlesquid minigame

nbouteme:
This adds a convenience option to simplify the battlesquid minigame, it does so by restoring the RNG state from a backup when the board is initialized, and updates the backup state only if the player sinks all the ships (squids?).

Basically the boats will be placed at the same positions until you win.


LagoLunatic:
I don't think I like this idea. Seems like a pointless option when players can choose to just not put progress items in battlesquid.
And since this minigame shows you the location of every squid when you lose, it completely removes the "game" element past the first time you encounter each board layout - you know 100% for sure what the next layout will be, so it's just a matter of remembering approximately where each squid was and then pressing the buttons necessary. At that point you might as well just cut out the middleman and give the player all the rewards whether they win or lose.

Also, there appears to be a bug with this code where the first board layout you encounter will always be this:
GZLE99-43
In other words you get that exact layout the first time you play battlesquid on every single seed, and also whenever you shut down the game and reboot it.

nbouteme:
I was watching a [100% race from Linkus](https://youtu.be/mQkRCRGcmd0?t=2276) and he mentionned this while failing for the 25th time. The thing is Linkus always checks all the progression location when generating a game, but he also has terrible luck.
You're right in that it makes it too easy, but the base minigame is too hard as playing optimally doesn't even guarantee winning, so I figured out it'd best to balance RNG and skill by making this more of a memory game.
Do you think it'd be decent if the solution wasn't shown when you lose and the player was limited to 12 or 8 bombs? I'll look into that.

Also the layout isn't exactly a bug, the initial seed backup is hardcoded. It can be either customized by the randomizer, but then it'll always be the same when restarting the game.
The other alternative was to backup the RNG code by modifying the NPC's REL to patch its loading code, but its code section is Yaz0-encoded and the tweak script doesn't seem to support that yet.

LagoLunatic:
>I was watching a 100% race from Linkus and he mentionned this while failing for the 25th time. The thing is Linkus always checks all the progression location when generating a game, but he also has terrible luck.
>You're right in that it makes it too easy, but the base minigame is too hard as playing optimally doesn't even guarantee winning, so I figured out it'd best to balance RNG and skill by making this more of a memory game.

I guess but all of the changes people have suggested making to this minigame to guarantee winning also fundamentally change the way it plays to the point that it's a different minigame. At that point I think you should just not play the minigame if you don't find it fun.

The whole purpose of the options that limit progress locations is so people aren't forced to do things they don't find fun after all. Many people don't find salvaging 49 sunken treasures fun but I haven't seen people suggesting that the sunken treasures be modified to salvage themselves automatically.

>Do you think it'd be decent if the solution wasn't shown when you lose and the player was limited to 12 or 8 bombs? I'll look into that.

Dunno, to me it just sounds more annoying really. With a reduced number of bombs it'll take you more attempts (meaning more rupees and more time skipping Salvatore's dialogue) to find the correct solution than it would in vanilla.

Also you're gonna have your work cut out for you if you intend to change the number of bombs, I tried that once and it's far from simple. For starters, the number of bomb icons displayed on the screen is part of the BLO1 UI layout file so you can't just remove them by reducing a number.

>The other alternative was to backup the RNG code by modifying the NPC's REL to patch its loading code, but its code section is Yaz0-encoded and the tweak script doesn't seem to support that yet.

get_raw_file automatically decompresses any Yaz0-compressed files (including RELs) that you request from it.
But if you're modifying a REL it's often easier to put it in an asm file than tweaks.py.

nbouteme:
I guess you're right, I'm closing this PR, then.

> But if you're modifying a REL it's often easier to put it in an asm

That's how I wanted to do it, but I'm not sure what address should I use in the org directive. Should it be the offset after decompressing, and are the RELs recompressed? I've seen a commit message saying compression wasn't used as it's too slow, which led me to believe it wasn't supported at all.

LagoLunatic:
>That's how I wanted to do it, but I'm not sure what address should I use in the org directive. Should it be the offset after decompressing, and are the RELs recompressed? I've seen a commit message saying compression wasn't used as it's too slow, which led me to believe it wasn't supported at all.

You use the offset when the REL is decompressed. If you run `py wwrando.py -disassemble` and hit randomize it will spit out all of the RELs decompressed as well as their disassembled code with symbols which is a lot easier to go through in a hex editor + text editor than by using dolphin.

That commit message is talking about recompressing files to save space. I finally got around to implementing that after like a year, but it turns out it's so slow that it would likely take hours to recompress everything whenever you generate a seed, so that's not happening. I still committed the code unused in case it helps with something later.

DEcompressing files that were compressed in the vanilla game has always been supported.

PR #136, opened 2019-05-14
Filter out most default items from spoiler log head

JarheadHME:


wwr_ui/randomizer_window.py
```diff
@@ -200,10 +228,30 @@ def randomize(self):
     self.ui.seed.setText(seed)
     self.update_settings()
     
+    important_options = [ # Options to track in spoiler log regardless of if they're default or not
+    "sword_mode",
+    "num_starting_triforce_shards",
```
wooferzfg: Why is `num_starting_triforce_shards` considered important but `randomize_entrances` isn't? Seems kinda arbitrary.

wwr_ui/randomizer_window.py
```diff
@@ -109,8 +109,8 @@ def __init__(self, cmd_line_args=OrderedDict()):
     # Hide unfinished options from the GUI (still accessible via settings.txt).
     self.ui.randomize_bgm.hide()
     
-    self.show()
-    
+    self.show()    
+
```
wooferzfg: You probably don't want to make changes to these two lines.

JarheadHME: That one was actually just an accident when testing stuff, but it didn't seem to break anything.

wwr_ui/randomizer_window.py
JarheadHME: Mainly because the default option for `randomize entrances` is "disabled" and I don't see this changing at any point in time, whereas `num_starting_triforce_shards` is just a number. Another would be because I asked lago about the starting shards and he said he thought it'd be better to keep that one in, but if that idea is ever changed, someone can feel free to change it since it's just a simple list.

LagoLunatic:
I don't think anything but starting items and custom colors need to be removed, randomize_entrances doesn't take up much space for example.
And it would be better to use a blacklist system than a whitelist so that options added in the future aren't removed by default, only options specifically listed would be removed.

Also, the function `update_total_progress_locations` also creates a dict of option values. For consistency to avoid possible bugs, both that function and actual randomization should use the same option dicts. So the code for generating a dict of options will need to be split into a separate function and that function called in both places.

JarheadHME:
How's this look for all of that?

PR #137, opened 2019-05-19
Make the seed depend on the build when the spoiler log isn't generated

wooferzfg:
Currently, it is very easy to cheat in races by running the randomizer from source, removing this line (https://github.com/LagoLunatic/wwrando/blob/master/randomizer.py#L367), and then looking at the spoiler log.

These changes make it so if the spoiler log isn't being generated (which probably means we're in the context of a race), the RNG depends on a key that gets generated when the executable is built. This prevents someone from being able to get the exact same seed when running from source unless they were the one who built the executable.

The key also gets saved to a file, which means that you can still debug issues with the randomizer that happened during races, as long as you were the one who built the executable. Changes to this file should obviously never be committed to GitHub.

These changes also add an encryption key when creating the executables with `PyInstaller`, which should make decompiling them more difficult (though not impossible, sadly).


LagoLunatic:
Does the key need to be generated every time a build is made? I'll need to manually keep track of a different key for every build I make that way, seems like a pain.

Also %random% in batch files depends on the current system time in seconds, so doing it this way would mean it's possible to reverse engineer what the key would be by simply looking at the modified at date of the executables I distribute and then just guessing how many seconds the build took and subtracting that. The range of how long a build could take isn't that large so you probably wouldn't have to guess that many times before you hit the right answer.

wooferzfg:
@LagoLunatic I made updates so that new keys only get generated when you run `generate_keys.bat`. I also got rid of `%random%`.

PR #148, opened 2019-07-27
Support PyQt5 and load UI without regenerating the Python code

ferzapata:
Although the randomizer runs without issue on Linux, the most common Qt bindings here are PyQt5, so this adds support for it as a fallback if PySide is missing.

Aside from that, the UI file is now loaded directly without having to regenerate the `ui_randomizer_window.py` file first, which is something that PySide can't do by itself for some reason.

This was tested on Arch Linux with both PySide and PyQt5 5.11.2.


LagoLunatic:
Unfortunately PyQt is GPL licensed, so I can't use it without relicensing WWRando as GPL too.

PySide seems like it supports Linux but users need to build it themselves. I don't use Linux myself but the instructions don't look complicated at all to me so IMO Linux users should just do that.

ferzapata:
Don't really need to build anything really, one just installs it from the package manager.

Il'l probably split the UI loading part into a new PR, since that still works fine without PyQt.

PR #149, opened 2019-07-27
Load UI dynamically without generated code

ferzapata:
The UI loading part without dependence on PyQt. Should make it so generating the python code from the Qt Designer UI file is no longer necessary.

Tested on Arch Linux with PySide 5.11.2, both from source and the Pyinstaller bundle.


LagoLunatic:
Honestly I'd much rather just click a batch file like normal than worry about all this weird eval and insecure xml parser stuff.

ferzapata:
…Seriously?

It's literally doing the same thing but in memory instead of on disk.

LagoLunatic:
If you say so. I'd rather just go with the regular code that comes with pyside.

I appreciate you trying to save me time but I'm personally not annoyed by running build_ui.bat, it only takes a second and I rarely have to run it.

ferzapata:
Well, in that case, you should at least be using the `pyside2-uic` tool that already comes with PySide, instead of the `build_ui.py` script. It'd be even less custom code.

LagoLunatic:
pyside2-uic only lets you compile one .ui file per call as far as I'm aware, not an entire folder at once.
While that is enough for wwrando since it only has one .ui file, I wrote the script case it was needed for generating a bunch of .ui files at once (which hasn't been useful for wwrando so far, but has for other projects).

ferzapata:
That's what the batch script should be for. No need to write your own tooling only to process more than one file.

```bat
@echo off
for %%f in (*.ui) do (
  pyside2-uic %%f > ui_%%~nf.py
)
```

There. That should generate the code for all the UI files in a directory.

LagoLunatic:
I'm not sure what you mean by "No need to write your own tooling". Your batch script and my python script are the exact same thing, just written in different languages.

I initially wrote it in python since I don't know batch, but I'm going to continue using the python one since it should have the benefit of working for people on other operating systems besides windows as well (haven't tested it but I assume it would work).

ferzapata:
I'd normally cite the benefit of the fact that batch/shell scripts can run without installing anything but the OS itself, but anyone who'd be using PySide would have Python installed, so you're right, guess it's fine as is.

But just for the record, if you wanted it to work pretty much anywhere (think it can even work in Windows, as of late) without depending on Python, you could use a shell script.

```sh
for f in *.ui; do
  pyside2-uic $f > ui_${f%%.ui}.py
done
```

But anyway, sorry for being an annoyance.

PR #151, opened 2019-08-26
Add option to start with custom amounts of health and double magic

JarheadHME:
Health is based on giving the player heart containers and heart pieces from the existing item pool.

Also adds support for spinbox widgets

Only existing problem that I'm aware of right now but seems to an underlying problem that at least I personally can't figure out how to fix, that being how on new game save files that were deleted with a certain amount of starting health are persistent to other isos until that file is deleted again.


LagoLunatic:
Merged in e3fa3b96320c87f8204842a5491ace5e2e1391df

PR #177, opened 2020-04-11
Cache the results of check_requirement_met

wooferzfg:
Tested with permalink: MS43LjBfYzA1Nzk2NABNZWFzbHlVbmtub3duV2hpcmxwb29sABcDBABOEMAoAgAAAAgAAA==

Dry run with old code: 2.44 seconds
Dry run with new code: 0.85 seconds

### DIFF
--- a/logic/logic.py
+++ b/logic/logic.py
@@ -63,6 +63,7 @@ class Logic:
     
     self.remaining_item_locations = list(self.item_locations.keys())
     self.prerandomization_item_locations = OrderedDict()
+    self.requirement_met_cache = {}
     
     self.done_item_locations = OrderedDict()
     for location_name in self.item_locations:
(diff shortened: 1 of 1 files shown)


LagoLunatic:
So the issue with this is that item requirement checking is pretty complicated, so it's very easy to forget to clear the cache for every possible thing that might change it. For example, you forgot to clear the cache when macros are changed dynamically, which results in logic bugs when the entrance randomizer is enabled.

Although it's not hard to fix the bugs once found, I'm not sure it's worth it to cache this and risk bugs when new features are added in the future since cache bugs can be annoying and the performance improvements are pretty small here. I generally only start worrying about optimizing if there's a way to save time noticeable even to the user like 30 seconds per seed or something.

wooferzfg:
Yeah, that's a fair point. I'll close this.

Issue #121, opened 2019-03-24
[Suggestion] Set custom number of required dungeons for Race Mode


TheHans255:
Right now, Race Mode is baked to put progress items in 4 dungeons, and the code for the items randomizer seems to have this set as a constant. Do you think we could get this as a settable range between 1 and 6?

For 5- and 6-dungeon races, there is currently a failsafe to add bow upgrades and the Hookshot in case the player started with all 8 Triforce Shards, and this should probably be extended to also add the Grappling Hook and Boomerang (both of which are logically required in Ganon's Tower, but aren't needed to complete Earth and Wind Temples or Forsaken Fortress) so that all six dungeon rewards can be filled in this case.

Are there any issues with 1- to 3-dungeon races that should be accounted for?

If you're OK with this change conceptually, I can implement it.

wooferzfg:
For 6 dungeon races, the worst case scenario is when you start with 8 shards and do swordless mode. You'd need to place 3 bow upgrades, Hookshot, Grappling Hook, and Boomerang.

For 1 dungeon races, the main issue is that the other 5 dungeons would be banned from having items. In dungeons only mode, this means very few total available progress locations.

Compynerd255:
> For 6 dungeon races, the worst case scenario is when you start with 8 shards and do swordless mode. You'd need to place 3 bow upgrades, Hookshot, Grappling Hook, and Boomerang.

I do mention the six dungeon case (in many words) in my original post. It is a little unfortunate that the Boomerang would have to be placed as a dungeon reward in this case (since it isn't technically required), but it could probably still be allowed since it isn't technically broken. And actually, since all six dungeons are required, a few more items also become required and are available for placement (such as Skull Hammer for FF/WT and Bombs for FF/TofG).

> For 1 dungeon races, the main issue is that the other 5 dungeons would be banned from having items. In dungeons only mode, this means very few total available progress locations.

Ooh, not to mention that the only required dungeon you could have in that case would be Dragon Roost Cavern, since every other dungeon has entrance requirements. In that case, we would either want to set the race mode minimum to 2 dungeons, or have smarter error handling, such as showing an error message to the user that the seed could not be generated, or placing the remaining unplaced items into the starting inventory.

Compynerd255:
[...]

PR #201, opened 2020-12-20
Add an option to choose the number of required dungeons in Race Mode

wooferzfg:
This PR adds an option to choose the number of required dungeons in Race Mode, allowing anywhere from 4 to 6 dungeons. I didn't include options lower than 4 dungeons because that causes too many errors in dungeons-only mode.

The Boomerang and Mirror Shield are now required as additional rewards when in Swordless mode and starting with 8 Triforce Shards. I chose these items because they're required to complete bosses but aren't required for any of the entrances.

This PR relates to #121, but it doesn't cover Race Mode with 1-3 dungeons as was mentioned in the issue.

### DIFF
--- a/randomizers/items.py
+++ b/randomizers/items.py
@@ -1,6 +1,7 @@
 
 import os
 import re
+from collections import OrderedDict
 
 from fs_helpers import *
 import tweaks
@@ -50,6 +51,13 @@ def randomize_items(self):
     self.logic.set_location_to_item(location_name, item_name)
 
 def randomize_boss_rewards(self):
+  # Try to generate dungeon boss reward locations until a valid set of locations is found.
+  for i in range(1, 50):
+    if try_randomize_boss_rewards(self):
+      return
+  raise Exception("Cannot randomize boss rewards! Please try randomizing with a different seed.")
+
+def try_randomize_boss_rewards(self):
   if not self.options.get("progression_dungeons"):
     raise Exception("Cannot randomize boss rewards when progress items are not allowed in dungeons.")
   
@@ -92,8 +100,6 @@ def randomize_boss_rewards(self):
       if item_name == "Progressive Bow"
     ]
     boss_reward_items += bow_upgrades[0:num_additional_rewards_needed]
-  
-  self.rng.shuffle(boss_reward_items)
 
   possible_additional_rewards = ["Hookshot", "Mirror Shield", "Boomerang"]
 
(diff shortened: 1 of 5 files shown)


LagoLunatic:
This gives a "Only items left to place are predetermined items at inaccessible locations!" error on this seed: MS44LjBfMDc0YjUyNQAyAAcBAwKOgMACAAAAAAAAAAA=
Same happens to most other seeds under these settings (rando entrances, 6 required dungeons, start with 8 shards, swordless).

LagoLunatic:
> This gives a "Only items left to place are predetermined items at inaccessible locations!" error on this seed: MS44LjBfMDc0YjUyNQAyAAcBAwKOgMACAAAAAAAAAAA=

Actually nevermind, looking at it closer this bug doesn't seem to be directly caused by this PR, it seems to occur when Hookshot is placed as the reward behind a rando'd entrance that requires Hookshot.

wooferzfg:
@LagoLunatic We could just add some logic when randomizing the boss rewards to not place the Hookshot in a dungeon that's behind one of the 3 entrances that require the Hookshot (Overlook, Pawprint Side Isle, Diamond Steppe). I'm not sure if that solution feels too hacky though, or if you have any suggestions for some other way to resolve this issue.

LagoLunatic:
I tried adding a hack to prevent the Hookshot issue which seems to work, but there are other similar issues which seem more difficult to fix. Cyclic dependencies, such as Jalhalla having Boomerang and Kalle Demos having Mirror Shield so you can't get either. Or 2/3 of the progressive bows being locked behind entrances where fire/ice arrows are required. These types of errors seem to happen in master too, but not nearly as frequently as with 6 required dungeons in this PR.

Not really sure what an effective fix for all of these would be. Maybe a way to detect if there's an impossible situation, and redo the boss reward selection in a loop until it gets a valid combination?

wooferzfg:
I added logic to validate the dungeon boss reward locations, and to regenerate those locations if they're invalid. This resolves most of the errors that come from cyclic dependencies.

However, there's still some seeds where the random entrances are so inconvenient that it's impossible to place all the dungeon boss rewards. I'm not really sure how to handle these cases. It seems to come up ~2% of the time with Swordless, 8 starting Triforce Shards, and all entrances randomized.

LagoLunatic:
I think a 2% failure rate under one specific set of options is fine for now, that's lower than the failure rate currently on master for those settings which was apparently low enough that no one had reported it as a bug.

PR #222, opened 2021-03-18
Add Seed Hash To Log Headers

tanjo3:
This PR adds writing the seed hash to log headers. This includes spoiler logs, non-spoiler logs, and error logs. The seed hash provides runners participating in a race an alternative method to permalinks for ensuring that they have the same seed. Having the seed hash in the logs allows the runners to perform this check without having to load up the game.

Another use for this is posting the seed hash in Racetime race lobbies. A bot can (already) be used to generate permalinks for races. With this PR, the bot would also be able to pull the seed hash from the non-spoiler log and update the race info with it. This would allow runners to check whether they have the correct seed without having to ask other runners in chat.

### DIFF
--- a/randomizer.py
+++ b/randomizer.py
@@ -78,6 +78,7 @@ class Randomizer:
     self.options = options
     self.seed = seed
     self.permalink = permalink
+    self.seed_hash = None
     
     self.dry_run = ("-dry" in cmd_line_args)
     self.disassemble = ("-disassemble" in cmd_line_args)
@@ -931,6 +932,9 @@ class Randomizer:
     if self.permalink:
       header += "Permalink: %s\n" % self.permalink
     
+    if self.seed_hash:
+      header += "Seed Hash: %s\n" % self.seed_hash
+
     header += "Seed: %s\n" % self.seed
     
     header += "Options selected:\n  "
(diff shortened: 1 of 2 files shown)


wooferzfg:
> A bot can (already) be used to generate permalinks for races. With this PR, the bot would also be able to pull the seed hash from the non-spoiler log and update the race info with it.

This is only true for spoiler log races. When the spoiler log isn't generated, the seed hash depends on the build key, which is embedded in the official randomizer release and is a secret.

tanjo3:
The build key is secret but the RNG seed which is used to generate the seed hash is still visible in the code. As such, the seed hash can be retrieved. I tested it when I did not generate a spoiler log and the seed hash in the non-spoiler log was correct. I don't see how it would be different in the official release.

wooferzfg:
The seed key is referenced here: https://github.com/LagoLunatic/wwrando/blob/e6688acec12d659d61ca11b2b092ac0a3f3b6bd5/tweaks.py#L1653

Since the bot runs the randomizer from source, it won't have access to the seed key, and therefore won't be able to retrieve the correct seed hash.

This won't be an issue with the official release though, so everything in this PR should work correctly.

LagoLunatic:
> I tested it when I did not generate a spoiler log and the seed hash in the non-spoiler log was correct.

Either you did accidentally both tests from source, or you just got the same seed hash by chance (1/3192 chance).

PR #230, opened 2021-04-24
Multi-platform releases and custom model installation

EthanArmbrust:
Builds release versions of the randomizer through GitHub Actions for Windows, Mac, and Linux with a release key (64 bit only). This depends on a repository secret called `RELEASE_KEY`.  
MacOS version is built as an app, with the program's settings and custom models folder located in `/Users/$USER/Library/Application Support/wwrando/`

This also includes a button on the Player Customization tab to install a custom model pack. This accepts a zip file containing a single model folder.  Checks are made for the existence of metadata.txt and Link.arc, with an error or success message for the appropriate case.

Some comments:
The macOS workflow uses miniconda instead of actions/setup-python to install python. This is because actions/setup-python builds python from source with the OSX deployment target set to the latest OSX version (10.15). This would mean that the resulting randomizer binary would not work on any macOS version older than 10.15. 

zipfile was replaced with shutil in `build.py`. This is because zipfile does not preserve the Unix executable flag when archiving directories.  

certifi package was added to `update_checker.py`. This was due to update checker throwing an SSL error in the binaries built by GitHub Actions. 

About Custom Models.txt will need to be updated to reflect the new installation method.


.github/workflows/python-app.yml
```diff
@@ -0,0 +1,106 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
```
wooferzfg, line 2: These comments don't seem relevant?

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,106 @@
...
+jobs:
+  linux:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: Add release key
+      run: echo ${{ secrets.RELEASE_KEY }} > keys/build_key.txt
```
wooferzfg, line 21: Builds started by the `pull_request` target don't have access to repository secrets (in order to prevent secrets from being stolen). Having a blank build key shouldn't cause any problems though.

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,106 @@
...
+    - name: Bundle Linux App
+      run: python build.py
+    - uses: actions/upload-artifact@v2
+      with:
+        name: wwrando-linux
+        path: dist/*.zip
+
+  macos:
+
+    runs-on: macos-latest
```
wooferzfg, line 43: It would be nice to be more consistent with the line breaks throughout this file.

wooferzfg (commented):
Just wondering:
- Have you tested the GitHub Actions workflows on your own fork? If so, do you have a link to the results?
- Have you tested out the resulting builds on Windows, Mac, and Linux?

EthanArmbrust:
> Have you tested the GitHub Actions workflows on your own fork? If so, do you have a link to the results?
Here's the run from the latest commit https://github.com/EthanArmbrust/wwrando/actions/runs/779877392
> Have you tested out the resulting builds on Windows, Mac, and Linux?
Yes, I've tested the binaries on my desktop running Windows 10 and Arch Linux 5.11.16, as well as a Macbook running macOS 10.13. I tested installing models as well as generating an iso.

.github/workflows/python-app.yml
EthanArmbrust: Oops, this is left over from the template and I forgot to remove it

LagoLunatic:
The button to install a custom model pack works for zips with a single model in it like this:
* zip
  * Name
    * Link.arc
    * metadata.txt

But does not work for [custom model packs](https://github.com/Sage-of-Mirrors/Custom-Wind-Waker-Player-Models) like this:
* zip
  * Custom-Wind-Waker-Player-Models-master
    * README.md
    * models
      * Name
        * Link.arc
        * metadata.txt
      * Name
        * Link.arc
        * metadata.txt

It would need to automatically detect which of the two structures is in use for a particular zip.

LagoLunatic:
Also, for the built releases, is it necessary for it to come in nested zip files?
* wwrando-windows.zip
  * Wind_Waker_Randomizer_1.9.0_64bit_win.zip

Can it just be Wind_Waker_Randomizer_1.9.0_64bit_win.zip by itself? Or is that just how GitHub Actions works?

LagoLunatic:
There's a visual bug that occurs when installing a model, probably fixable by not switching to the new model until *after* the user clicks OK:
image

EthanArmbrust:
Here's a 

> Also, for the built releases, is it necessary for it to come in nested zip files?
> 
>     * wwrando-windows.zip
>       
>       * Wind_Waker_Randomizer_1.9.0_64bit_win.zip
> 
> 
> Can it just be Wind_Waker_Randomizer_1.9.0_64bit_win.zip by itself? Or is that just how GitHub Actions works?

Unfortunately, this is a limitation of the `upload-artifact` action. https://github.com/actions/upload-artifact#zipped-artifact-downloads

.github/workflows/python-app.yml
EthanArmbrust: Should be better now

EthanArmbrust:
> It would need to automatically detect which of the two structures is in use for a particular zip.

I've updated `install_custom_model_zip` to support custom model pack zip files, as well as single model zip files. I've also fixed that visual bug.

wooferzfg:
> Unfortunately, this is a limitation of the upload-artifact action. https://github.com/actions/upload-artifact#zipped-artifact-downloads

Rather than uploading a ZIP file, I think you could just upload a folder and allow the upload-artifact action to zip it for you.

wooferzfg:
@EthanArmbrust How is `keys/seed_key.py` set after these changes?

wooferzfg:
@LagoLunatic I think you'll need to add `RELEASE_KEY` as a secret to this repository, with its value matching the value of `keys/build_key.txt` that you have locally.

LagoLunatic:
> @LagoLunatic I think you'll need to add `RELEASE_KEY` as a secret to this repository, with its value matching the value of `keys/build_key.txt` that you have locally.

Done, it builds now.

> @EthanArmbrust How is `keys/seed_key.py` set after these changes?

Seems like the file is not being created so it defaults to using no seed key.

wooferzfg:
I think it would be good to add a `SEED_KEY` secret and create the file based on that.

LagoLunatic:
Done.

@EthanArmbrust Can you check the linux and mac versions to be sure they're using the seed key correctly? (Or anyone else with access to mac and linux)

[Download the latest build](https://nightly.link/LagoLunatic/wwrando/workflows/python-app/master), don't touch any settings except setting the permalink to `MS45LjAAQWZyYWlkVGltZWxlc3NCYWl0bwAHAQMADiBAAwAAAAAAAQAA`, and check if the resulting ISO has the MD5 hash `4efad906a649973471da7e71663d6980`.

EthanArmbrust:
Mac is working and has matching md5 hash.
Linux is seg faulting when running from release binary, I'll investigate the cause.

EthanArmbrust:
Just tried on a live Ubuntu USB, it seems to be working and the md5 hash also matches

PR #240, opened 2021-06-12
Add new convenience tweak: Invert sea compass x axis

fig02:
This flips the sea compass on the x axis so it can more accurately reflect the direction the player is facing.
This is technically less accurate according to how compasses actually work, but this is still more intuitive for some people.

This was a fairly simple change, it nops one `neg` instruction without adding any additional code.
Specifically, it changes `-((pfVar2->mCollisionRot).y - 0x8000)` to `((pfVar2->mCollisionRot).y - 0x8000)` in `dMeter_compassGetOnProc`.

image1

An option was added to the GUI under `Convenience Tweaks`

image2

This change is something myself and a couple friends wanted for our own personal use, but I figured I should open a pull request since others may find it desirable too.

I'm not too active in the TWW randomizer community, but I've done plenty of work on the OoT randomizer (and OoT decomp).
Learning about this game to make this change was fun! Hope to learn more.


wooferzfg:
The new option should probably be in the `NON_PERMALINK_OPTIONS`: https://github.com/LagoLunatic/wwrando/blob/9b00a447619677d2c5c95d34be86c84096be60b9/wwr_ui/options.py#L225

fig02:
done, thanks woofer

PR #265, opened 2022-02-15
Chest type matches contents

tanjo3:
_Chest Type Matches Contents_, or **_CTMC_** for short, is a setting that changes the type of a chest to match the item it contains. The original game's chests come in four types: light wood, dark wood, metal, and Big Key. CTMC uses the first three types of chests, as Big Key chests are not cooperative with this setting. Since the two types of wooden chests can be challenging to distinguish, we replace the texture of the dark wood chests with a custom texture, seen below (middle). This texture borrows from the color scheme of the Big Key chests, for reasons we explain below. Accordingly, CTMC modifies the chest type to either a wooden chest, a key chest, or metal chest.

Chest Types in CTMC

We consider the items in the game in three categories. Progress items--items required to access progress locations in the seed--will be placed in metal chests. Dungeon keys are always considered separate from progress items (even in key-lunacy) and will be placed in key chests. Finally, the remaining items (i.e., non-progress items and consumables) will be placed in wooden chests. Keys for non-required dungeons are considered as non-progress items and will also be found in wooden chests.

This setting adds a new way to play and experience the randomizer. While not all checks in the game are chests, seeing a chest and knowing what type of item it is can speed up the game. At the same time, if you never open wooden chests, you will likely be low on hearts and Rupees. Thus, there is a cost-benefit tradeoff to which chests you open for casual play. Additionally, many chests can be seen from afar without actually being able to access them immediately. In these cases, the player can know whether they need to continue with unlocking access to that chest or if they need to return and make a second visit later. Again, this can speed up the game and help with how the game flows.

CTMC also has competitive implications. The racing community has been testing racing with CTMC, and the general sentiment is overwhelmingly positive. Speeding up seeds allows for races with standard settings to be run quickly, increasing the accessibility of racing the randomizer. It could also open up the possibility of using seeds with more settings. In addition, seeing chests from afar (also known as peeking) reduces double dips to islands and dungeons, making seeds feel smoother to run. Strategies to peek chests becomes a new mechanic to races; a document overviewing these peeks can be found here: http://bombch.us/DRup.

Overall, CTMC adds an extra check to set the chest type at the same time the randomizer places the randomized item into the chest. Additionally, since the chest sizes are nearly identical between chest types, it doesn't affect how the chests behave in the game (which is why we don't use the Big Key chest type). The details of CTMC have been a collaborative effort by the randomizer racing community, and we would enjoy sharing this with the general WWR audience.


wooferzfg:
Could you rebase this onto the `master` branch of `LagoLunatic/wwrando` and fix the merge conflicts? (You can DM me if you have any questions about how to do that.)

PR #272, opened 2022-05-03
Add Old Man Ho Ho as an alternative hint placement

tanjo3:
This PR adds the Old Man Ho Ho as an alternative hint placement to the fishmen.

<p align="center">

</p>

Fishmen hints, at least in a competitive context, are underutilized. This is generally because the cost-benefit of getting hints is low. Fishmen hints require the Bait Bag, and the player needs to go out of the way to purchase All-Purpose Bait to feed the fishmen. By the time the player can do so, they might have already completed a good bit of the seed. As such, since each fishmen only gives one hint, there is a good chance that the hint will be for an item that was already found. Finally, because at least three copies of each hint exist, the chance of finding a replicate of a hint you have already found is high.

Old Man Ho Ho hints attempt to address some of these issues. There are 10 Old Man Ho Ho in the world. Eight of them require no items to get to; the Old Man Ho Ho on Horseshoe Island and Outset Island requires the Deku Leaf at the minimum. In order to distribute the default 15 item hints evenly, each hint is duplicated, and each Old Man Ho Ho is given three hints. Thus, getting hints earlier in the seed is more accessible and stronger. Furthermore, thematically, Old Man Ho Ho already gives hints in the vanilla game, so this hint placement makes logical sense.

This PR adds an option to the UI to place the hints on Old Man Ho Ho. This setting can be use in conjunction with placing hints on the fishmen, in which case the total amount of hints is split evenly between the two placement options. The user can also specify the total number of hints. If they would like to remove hints from the seed, they can leave both placement options unchecked. Note that the hint placement options do not affect the Wallet mail, Big Octo Great Fairy, or Savage Labyrinth hints, and they remain unaffected by this PR. To generalize and organize hint-related code, this PR creates a `hints.py` file that can be easily maintained and modified in future hint changes.

In the vanilla game, the Old Man Ho Ho on Outset Island, Two-Eye Reef, and Private Oasis disappear after a particular flag is set (the stone head on Savage Labyrinth is removed, or the Big Octo in the quadrant is defeated). Since Old Man Ho Ho is now tied to hints, an ASM patch provided by SuperDude88YT#3196 prevents these Old Man Ho Ho from despawning. Credit also to kinnian#7186 for reporting this during testing. Finally, the distribution method for Old Man Ho Ho hints generalizes to a variable number of hints, not just the default 15. First, the number of hint replications is computed to be able to divide the hints evenly. Then, the hints are randomly assigned to the Old Man Ho Ho. One hundred attempts are made to not have any single Old Man Ho Ho give the same hint more than once to make each visit more beneficial to the player. However, if no such condition is met within the hundred attempts, the hints will just be assigned to the Old Man Ho Ho ignoring hint replication.

~This PR also fixes a rando bug where hints for the Pawprint Isle Wizzrobe Cave would be incorrectly tied to the Pawprint Isle Chuchu Cave, which would result in incorrect hints when secret cave entrance rando is turned on.~ This issue is resolved by #282 and the PR has been updated to reflect these changes.

Edit: This PR also changes the seed creation for the randomizer's RNG. In order to prevent cheating, we want to have the hint options affect how the seed is randomized. Currently, it would only affect the seed hash. In this change, when the RNG is reset, additional `randbits()` calls are made depending on which hint-related options are enabled.


wooferzfg:
As a casual player, I might want to have both the Fishmen and the Old Man Ho Hos give hints. In fact, that might even make sense as the default setting. Would it make sense to have a "Hints" section or tab with a checkbox for each type of hint location instead of a combo box?

tanjo3:
Hmm, interesting. I'll look into it. Not sure how to exactly handle the replication of hints when both Fishmen and Old Man Ho Ho have hints.

wooferzfg:
I would choose some number of total hints to generate, and then split them evenly between the Ho Hos and the Fishmen. (For example, if we want 20 hints total, then 10 would go to the Fishmen, and 10 would go to the Ho Hos, and then those respective methods would assign duplicate hints as necessary.)

LagoLunatic:
There seems to be an issue where Hoho keeps saying the same hints over and over and over again, instead of just saying one or two unique hints. Here's one example of a Hoho's dialogue:
```
[sound=259]Ho ho! They say that [red]a piece of 
the power of the gods[white] is located 
in [red]a very populated place[white][wait=30]

[draw_instant]and that [red]an explosive item[white] is 
located in [red]the home of the wind[white][wait=30]


[draw_instant]and that [red]an item that always comes 
back to you[white] is located in [red]a place 
where the vegetation is abundant[white][wait=30]

[draw_instant]and that [red]an item that always comes 
back to you[white] is located in [red]a place 
where the vegetation is abundant[white][wait=30]

[draw_instant]and that [red]an upgrade for your blade[white] 
is located in [red]your island[white][wait=30]


[draw_instant]and that [red]an upgrade for your blade[white] 
is located in [red]your island[white][wait=30]


[draw_instant]and that [red]an upgrade for your blade[white] 
is located in [red]the hottest place in 
the Great Sea[white].[wait=30]
```
That's all just one single Hoho's dialogue.

The other Hohos in the same seed also repeat hints over and over, though exactly which are repeated is slightly different for each Hoho.

Settings were default except I enabled hoho hints: MS45LjBfMzBkMDg5MwBFbmVteVJhbmRvVGVzdAAHAQMA7A4AQAMAAAAAAAAQAAA=

LagoLunatic:
Also, there's a potential bug in the implementation of RNG_CHANGING_OPTIONS.

It tries to make the number of times it changes the RNG seed unique based on the combination of options selected with `100 * self.options.get(option) * (1 << i)`.
This works fine assuming all boolean (0 or 1) options, but since num_hints is an integer, a cheater can theoretically set it to specific values to simulate other combinations of options later on in the list.

For example, if a race is intended to be played with num_hints=0 and do_not_generate_spoiler_log=True, a cheater could set num_hints=2 and do_not_generate_spoiler_log=False, and get the exact same random seed, despite a different combination of RNG_CHANGING_OPTIONS. This would allow full access to the spoiler log and 2 hints, when there's supposed to be no spoiler log and 0 hints.
This is because `(100 * 1 * (1 << 3))` == `(100 * 2 * (1 << 2))` == 800.

This particular example doesn't currently work because the minimum you can set num_hints to is 1, not 0. But it would be easy for a bug like this to be introduced in the future if how num_hints works is changed, or if more options are added to RNG_CHANGING_OPTIONS, so it might be safer to come up with an uncheatable implementation now rather than worrying about this introducing bugs in the future.

tanjo3:
For the repeated hints, this seems to be intended behavior. It looks strange because, with those settings, there are 7 hints assigned to the Old Man Ho Ho. In order to distribute 7 hints evenly across 10 Old Man Ho Ho, each hint must be repeated 7 times. The way the distribution among Old Man Ho Ho works, it will likely fail to find a distribution without repetition on any single Old Man Ho Ho after `n_attempts` attempts, since 7 hints per Old Man Ho Ho is an edge case where each Old Man Ho Ho would need to have a copy of every single hint.

To address this, I've modified the failsafe behavior for hint distribution to iteratively distribute the hints among the Old Man Ho Ho in the event `n_attempts` distribution attempts have been made without success.

LagoLunatic:
It fixed each Hoho repeating himself, but they still give way too many hints, with every Hoho saying every hint, so all hints are duplicated 10x throughout the game:

```
[sound=259]Ho ho! They say that [red]a piece of 
the power of the gods[white] is located 
in [red]a very populated place[white][wait=30]

[draw_instant]and that [red]a piece of the power of 
the gods[white] is located in [red]the rubble 
of the evil's lair[white][wait=30]

[draw_instant]and that [red]a piece of the power of 
the gods[white] is located in [red]the home of 
bird people[white][wait=30]

[draw_instant]and that [red]an explosive item[white] is 
located in [red]the home of the wind[white][wait=30]


[draw_instant]and that [red]an item that always comes 
back to you[white] is located in [red]a place 
where the vegetation is abundant[white][wait=30]

[draw_instant]and that [red]an upgrade for your blade[white] 
is located in [red]your island[white][wait=30]


[draw_instant]and that [red]an upgrade for your blade[white] 
is located in [red]the hottest place in 
the Great Sea[white].[wait=30]
```

Why does every Hoho need 7 hints under default settings? Each Hoho could even say as many as 10 hints depending on what the user sets num_hints to. You said that each Hoho is supposed to have 3 hints, which already seems a bit high to me, but any more than 3 is *really* overkill.

It seems like all of this is a result of trying to make sure that every Hoho is even with every other one, but I think I'm missing why that matters. Couldn't some Hohos have 2 hints, some have 3, some have 1, etc? And if there are fewer hints than Hohos, a few duplicates can be made to make sure they each have at least 1 hint instead of 0.

tanjo3:
The fact that each Old Man Ho Ho, under most settings, have more than one hint is what makes them more appealing to visit than fishmen. The intent of this PR is to make hints a more attractive option, which in my opinion, the current fishmen hints are not. Furthermore, I'd like to see hints be more competitively viable, which is why having each Old Man Ho Ho having the same number of hints is important. In tournament settings, not all Old Man Ho Ho will be conveniently-accessible, and going to a less-convenient Old Man Ho Ho and only getting one hint feels really bad.

I'd agree that a single Old Man Ho Ho having 7 hints and each Old Man Ho Ho saying the exact same thing is a bit strange. I can make a change where hints are replicated until there are a multiple of 10. So, in this case, there'd be 7 hints + 3 replicated for a total of one hint per Old Man Ho Ho. And for the default number of 15 hints (and not placing any hints on the fishmen), each Old Man Ho Ho would have 2.

LagoLunatic:
> This PR also fixes a rando bug where hints for the Pawprint Isle Wizzrobe Cave would be incorrectly tied to the Pawprint Isle Chuchu Cave, which would result in incorrect hints when secret cave entrance rando is turned on.

I'm not sure that this fix is correct. You use the zone name "Pawprint Isle Side Isle" in this fix, but there is no such zone name in item_locations.txt. I think that making this change would just cause hints for the Wizzrobe Cave to never be placed.

This should probably be a separate issue though.

tanjo3:
Ah, yeah. The behavior is correct when entrances are randomized, but not when they are vanilla. I can make a separate PR with the changes,

LagoLunatic:
Can this PR still be rebased to get rid of all the duplicate commits from master? Not too familiar with rebasing.

tanjo3:
OK, I am obviously making it worse lol

tanjo3:
Alright, so I forgot to force push the first and second time I rebased. Not sure how to fix that.

LagoLunatic:
I think I have it fixed locally but github isn't letting me force push.
Is this the right command or am I doing it wrong? `git push origin hoho-hints -f`

tanjo3:
That looks correct for a force push to me, but I wouldn't trust me on it.

LagoLunatic:
Does it look correct now?

(The issue I was having force pushing from the command line was resolved by [this](https://stackoverflow.com/a/68781050) for future reference.)

tanjo3:
Yeah, everything seems to be in order. Thanks for fixing that; I'm still very bad at git stuff and am learning through working on the randomizer bit by bit.

There's a bug where an error is thrown when no hint placement options are selected. I don't recall if I addressed this in previous commits, but I can make a fix for it now.

PR #283, opened 2022-06-12
Add King of Red Lions as a hint placement option

tanjo3:
This PR adds King of Red Lions as an alternative hint placement option.

The current hint placement options available to the player are the fishmen and Old Man Ho Ho. Both these options share that the player must go out of their way to get these hints, sometimes extremely so. For example, the player might get the Bait Bag for fishmen hints late into the seed, lessening the usefulness of hints. Additionally, these hint placement options can be tricky to use competitively. For example, one runner might go far out of their way to talk to an Old Man Ho Ho but get rewarded with no helpful information. Or the opposite could happen where this risky venture gives the runner valuable information as to where an item like Light Arrows are. Some might find that desirable, but I would argue that a more fair hint placement option should be available.

As a hint placement option, King of Red Lions (KoRL) addresses these issues because the player can talk to KoRL at almost any point during the seed. Additionally, since KoRL gives all the hints, all runners in a race are working with the same information at the same time. This means that hints can be used to guide the player's routing without introducing the routing risk of actually getting the hints. Of course, how a player uses the information from hints introduces risk. Still, I would argue that this is the intended application of having hints in a randomizer.

Since KoRL tells the player all the hints, he has one textbox for every hint assigned to the hint placement option. Therefore, we remove the delay for KoRL's textboxes, typically added to hint text to stop the player from passing over a hint. This change reflects community feedback on KoRL hints. It lessens the time loss accrued when accidentally talking to KoRL, which can happen multiple times per seed. Additionally, since talking to KoRL is very accessible, if a player does skip over a hint, talking to KoRL again is not punishing.

### DIFF
--- a/hints.py
+++ b/hints.py
@@ -31,10 +31,11 @@ def generate_item_hints(self, num_hints):
   octo_fairy_hint = None
   fishmen_hints = []
   hoho_hints = []
+  korl_hints = []
   unique_hints = []
   
   # Identify where the user wishes hints to be located
-  variable_hint_placement_options = ["fishmen_hints", "hoho_hints"]
+  variable_hint_placement_options = ["fishmen_hints", "hoho_hints", "korl_hints"]
   num_hint_placements = sum(self.options.get(option) for option in variable_hint_placement_options)
   
   # Always assign one hint to the Big Octo Great Fairy
@@ -121,6 +122,8 @@ def generate_item_hints(self, num_hints):
       fishmen_hints.append(item_hint)
     elif current_hint_placement == "hoho_hints":
       hoho_hints.append(item_hint)
+    elif current_hint_placement == "korl_hints":
+      korl_hints.append(item_hint)
     
     # Move the next hint placement
     hints_remaining_per_placement[current_hint_placement] -= 1
@@ -143,11 +146,18 @@ def generate_item_hints(self, num_hints):
     # Duplicate a hint that we've already made.
     hoho_hints.append(self.rng.choice([octo_fairy_hint] + fishmen_hints))
   
+  if self.options.get("korl_hints") and len(korl_hints) == 0:
+    # Unable to make a hint for King of Red Lions, but was able to make one for the Big Octo Great Fairy and the fishmen.
+    # Duplicate a hint that we've already made.
+    korl_hints.append(self.rng.choice([octo_fairy_hint] + fishmen_hints))
+  
   # Loop through the hints, converting the item and location names to hint strings
   octo_fairy_hint = (self.progress_item_hints[octo_fairy_hint[0]], self.island_name_hints[octo_fairy_hint[1]])
   for i, hint in enumerate(fishmen_hints):
     fishmen_hints[i] = (self.progress_item_hints[hint[0]], self.island_name_hints[hint[1]])
   for i, hint in enumerate(hoho_hints):
     hoho_hints[i] = (self.progress_item_hints[hint[0]], self.island_name_hints[hint[1]])
+  for i, hint in enumerate(korl_hints):
+    korl_hints[i] = (self.progress_item_hints[hint[0]], self.island_name_hints[hint[1]])
   
-  return octo_fairy_hint, fishmen_hints, hoho_hints
+  return octo_fairy_hint, fishmen_hints, hoho_hints, korl_hints
--- a/randomizer.py
+++ b/randomizer.py
@@ -70,6 +70,7 @@ CLEAN_WIND_WAKER_ISO_MD5 = 0xd8e4d45af2032a081a0f446384e9261b
 RNG_CHANGING_OPTIONS = [
   "fishmen_hints",
   "hoho_hints",
+  "korl_hints",
   "num_hints",
   "do_not_generate_spoiler_log",
 ]
--- a/tweaks.py
+++ b/tweaks.py
@@ -887,7 +887,7 @@ def randomize_and_update_hints(self):
   num_hints = int(self.options.get("num_hints", 0))
   
   # Generate item hints
-  octo_fairy_hint, fishmen_hints, hoho_hints = hints.generate_item_hints(self, 1+num_hints)
+  octo_fairy_hint, fishmen_hints, hoho_hints, korl_hints = hints.generate_item_hints(self, 1+num_hints)
   
   # Give the Big Octo Great Fairy a unique item hint
   update_big_octo_great_fairy_item_name_hint(self, octo_fairy_hint)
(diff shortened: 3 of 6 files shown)


LagoLunatic:
I don't like the idea that KoRL dumps a million hints on the player right at the start. Feels like it's spoiling too much for the player all at once for no work and with no exploration, not really in the spirit of a randomizer. And even if it's not a default option, lots of people will just check every option available, and newer players will probably talk to KoRL before exploring.

I think KoRL should either be limited to 1 hint per run total, or have his hints split up throughout milestones in a run. I'm not sure exactly how those milestones should work, but maybe one hint before beating any dungeon bosses, a second hint after defeating one dungeon boss, third after two, etc? So you would only be able to view one hint at any given time. Or it could be the number of triforce shards you've obtained instead of bosses defeated maybe.

tanjo3:
It's an understandable concern about the information dump at the start of a seed that KoRL hints might bring. However, I would say that item hints exacerbate this issue. With other potential hint types (which I plan on bringing up to PR sometime in the future), the hints KoRL gives would not necessarily be as strong.

Furthermore, I cannot entirely agree with the "spirit of a randomizer" argument because different people enjoy randomizers differently. While some enjoy the freedom of exploration, others wouldn't mind more direction. As I mentioned, KoRL hints are, in a sense, fairer in a competitive setting than other hint placement options. Many people enjoy the randomizer by racing it and would agree with this change. To back up that statement, we recently had a survey to decide on the settings for the S5 Tournament. Over 75% of respondents who agreed with having a form of hints in the tournament also agreed with having the hints on KoRL.

Having an alternative implementation as you outline would be good as well. However, there are numerous issues (as you say), such as how to decide on milestones, how to implement that system, and the fact that you would lose access to a hint after completing a milestone. I agree that such a system would be conducive to casual play. However, I don't think hints' utility should be limited only to casual play.

To address your concerns, we could have KoRL hints as a hidden option, just like the option to require Tingle Bombs for Tingle chests. This way, competitive runners can still use KoRL hints, while newer runners wouldn't accidentally turn KoRL hints on. Alternatively, we could make KoRL hints only an option when Race Mode is turned on. Race Mode is already an option geared toward competitive play, so pairing KoRL hints with this setting only would make sense.

In summary, while your concerns about how KoRL hints would change how the randomizer is played are justified, I think it is necessary to consider all audiences. The randomizer should not be geared solely to casual play. Considerations should be made to those who prefer to enjoy the randomizer competitively, and hints do not address this as they are now. To force the use of KoRL hints would be wrong, but to disavow the competitive utility of KoRL hints would also be wrong.

LagoLunatic:
> It's an understandable concern about the information dump at the start of a seed that KoRL hints might bring. However, I would say that item hints exacerbate this issue. With other potential hint types (which I plan on bringing up to PR sometime in the future), the hints KoRL gives would not necessarily be as strong.

How does adding more options for where hints can appear make one of the existing ones less strong?
Or do you mean changing KoRL to hint at something else randomized besides items, like which entrances lead to what dungeons/caves? Because I feel like KoRL hinting at more than one entrance would be even worse, as there are fewer entrances than items.

> Having an alternative implementation as you outline would be good as well. However, there are numerous issues (as you say), such as how to decide on milestones, how to implement that system, and the fact that you would lose access to a hint after completing a milestone. I agree that such a system would be conducive to casual play. However, I don't think hints' utility should be limited only to casual play.

How to implement it would be pretty simple, KoRL's init message could be set to the base message ID + the return value from a function like `getTriforceNum` or a custom function that counts dead bosses so that the message changes as the game progresses.

Losing access to the hint would be an issue though, you're right. A system that makes sure hints only disappear once you've gotten the item would be a lot more complicated to code.

> To address your concerns, we could have KoRL hints as a hidden option, just like the option to require Tingle Bombs for Tingle chests. This way, competitive runners can still use KoRL hints, while newer runners wouldn't accidentally turn KoRL hints on.

I don't like the existence of the hidden tingle bomb option either and I'm planning to remove it since I don't think anyone uses it anyway. Or is that actually used by competitive runners?

> Alternatively, we could make KoRL hints only an option when Race Mode is turned on. Race Mode is already an option geared toward competitive play, so pairing KoRL hints with this setting only would make sense.

Hmm, that might make sense to do. Though I guess there might be cases where KoRL hints are desirable when race mode is off?

Or instead, maybe all the hint options (+no spoiler log option) should be moved into a separate "Advanced Options" tab? Casual players probably don't need to adjust the hint settings if the defaults are good but the option would still be there for those who need it.
I was thinking about adding another tab anyway since the randomizer window has grown too tall with the hint options, to the point that [11.53% of people's primary monitor](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam) can't fit all the options on screen without scrolling.

wooferzfg:
> Or is that actually used by competitive runners?

It's not really used. It should be safe to remove it.

Also, would it make sense to move the options related to Race Mode to the Advanced Options as well?

LagoLunatic:
> Also, would it make sense to move the options related to Race Mode to the Advanced Options as well?

That probably makes sense come to think of it.

tanjo3:
> How does adding more options for where hints can appear make one of the existing ones less strong?
Or do you mean changing KoRL to hint at something else randomized besides items, like which entrances lead to what dungeons/caves? Because I feel like KoRL hinting at more than one entrance would be even worse, as there are fewer entrances than items.

Sorry, I wasn't clear. I don't mean adding more hint placement options. Rather, I mean different hint types in addition to hints pointing to directly to progress items. Once I set up a PR, this will hopefully be more clear, but without that context, I understand what I'm trying to say can be unclear. An example would be a hint that tells you what is on a particular check, even if the check has nothing of value. So knowing that something like Pawprint Chuchu Cave - Scale the Wall is just a Golden Feather, for instance. Another example would be that something on Outset Island is important. Similar to item hints, you don't know what check the hint is talking about, but you also wouldn't know what item it is. All you would know is that it's important.

Even with item hints on KoRL, the game doesn't necessarily become "go here, do this, go here, do this", but with these additional types of hints, you have to plan out your route even more to work around the information that you're given. It's this kind of strategy and logic reading that is well-received by many people in the racing community, and even in casual play.

> Or instead, maybe all the hint options (+no spoiler log option) should be moved into a separate "Advanced Options" tab? Casual players probably don't need to adjust the hint settings if the defaults are good but the option would still be there for those who need it.

I think that this solution would work. If you'd like, I can make changes to add such a tab in this PR. Or, I could do a separate PR with this change, then rebase this PR onto that. Let me know what you think.

LagoLunatic:
> I think that this solution would work. If you'd like, I can make changes to add such a tab in this PR. Or, I could do a separate PR with this change, then rebase this PR onto that. Let me know what you think.

You can add them to this PR if you want.

tanjo3:
<p align="center">


</p>
Added the new tab, moved things around a bit, and resized the window. Let me know what you think!

PR #286, opened 2022-07-31
Small improvements

DualVission:
Softened the key chest texture
Moved sea minigame jingles to long non-looping jingles
Improved enemy palette randomization (Py 3.10 req)


LagoLunatic:
1. Please revert all the changes to indentation you made.
2. The rando currently uses Python 3.9, the workflows haven't been updated to build with 3.10 yet. I will do this eventually but the PR cannot be merged until then, not sure when.
3. What was the specific reasoning behind the change to enemy palette randomization? Do you have a before/after screenshot?

LagoLunatic:
Also, why are the enemy names in the case statement duplicated in an if statement beforehand? That actually makes the code worse than before the addition of the case statement.

DualVission:
Wait, why is this on the main branch? I opened this pull from Tanjo's branch?

LagoLunatic:
Your fork only has a single branch named "master". There's no other branch you could have opened it from.

PR #287, opened 2022-08-01
Improvements to Music Constants and Key Chest

DualVission:
Sea minigame jingle is too long to be enjoyable on items more than once unless we can figure out why chart 23 is special 
Chest Before 
image 
Chest After* The keyhole has been darkened since this screenshot 
image 
 
Can you tell I don't use GitHub enough?


LagoLunatic:
> chart 23 is special

In what way is it special?

---

Also for the before/after, I meant for the change you made to enemy palette randomization. The `abs()` calls. What are those doing?

DualVission:
`abs()` would prevent passing negative numbers to the HSV adjustment calculation without skipping an enemy. I've never personally seen it skip an enemy or texture, so comparing and contrasting before and after is not necessarily the easiest thing to pinpoint. Another option would be replacing the `continue` with `r, g, b = abs(r), abs(g), abs(b)`
As for chart 23, it seems with music rando that it will allow you to exit dialog sooner than the end of the jingle. Neither tanjo3 nor I have found why it is special, just that it is. I have only done limited investigation.

LagoLunatic:
Oh, I thought you actually knew what materials had negative values, not that you were just changing things at random.

I don't think adding `abs` there makes sense, we don't even know WHY the values are negative in the first place. That's why I skipped them. A proper solution would be to look into how the game engine/gamecube hardware (or dolphin's code) handles negative values there, and then make a decision about how to randomize those colors based off of that information.

DualVission:
It's not at random, I do agree with the approach though with the "skip because we're not sure what it does". In my experience with shading (in this very PBR contemporary era), negative values are only the result of bad math or errors (the only notable exception is IOR, but that is a conversation for another time).

LagoLunatic:
The changes to palettes.py still don't make sense, you just added the same code commented out and updated the existing comment to be wrong.

Also I'm not planning to merge any PRs that tweak music rando since the whole thing is broken and may as well be removed until it can be rewritten from scratch in a different way. No point in working on it as-is.

The chest texture does look better, revert everything else you changed if you want me to merge that one.

PR #293, opened 2022-09-18
Removing a Code Smell

Crain-32:
Noticed the reassignment while adding in type hints on my own Fork.


wooferzfg:
I don't actually think this causes any problems, but it seems to be bad practice. See [this post](https://stackoverflow.com/a/65507051).

PR #295, opened 2022-09-25
Enemy randomizer - difficulty settings

zacharyg1:
Enemy randomizer often results in very difficult seeds due to placing combinations of enemies that weren't intended to be together in the main game. This pull request adds a "Difficulty" option to the enemy randomizer, to prevent hard enemies from spawning, or to make hard enemies spawn more.

The difficulties are:
- Fully random enemies: How the enemy randomizer works currently. Enemies are assigned purely at random.
- Easy enemies: Most enemies will be one-shots (keese, rats) or killed with one combo (bokoblins)
- Medium enemies: A combination of easy enemies and ones that require more effort, such as moblins/armos
- Hard enemies: Excludes the one-shot enemies, and adds enemies that are a serious threat such as Stalfos and ReDeads
- Very Hard enemies: The same as Hard, but includes enemies that can be unreasonable in large groups (Darknuts, Octorok Spawners), can cause softlock (Blue Bubble, Dexivine), or are extremely punishing of mistakes (Floormasters)

When any difficulty except "Fully Random Enemies" is enabled, the randomizer will not allow you to play with Spoils Trading progress items on, since certain enemies won't spawn on some difficulties, making farming spoils items impossible.

Flying Mothula are excluded from all difficulties and only spawned when playing on Fully Random Enemies mode. When Flying Mothulas are added to Very Hard, the randomizer often fails to generate a seed due to how much memory Flying Mothulas consume, together with not having low-memory enemies like Keese available to fill in the remainder.

Additionally, Flying Mothulas have infinite follow range which can cause problems in some places. They're also just frustrating to play against without Fire Arrows.

This PR also fixes a minor typo in the Enemy Randomizer data - subequent -> subsequent

### DIFF
--- a/data/enemy_types.txt
+++ b/data/enemy_types.txt
@@ -298,7 +298,7 @@
   Can be killed with thrown objects: Yes # TODO CHECK
   Can be killed with bomb flowers: Yes
   # Note: Cannot escape from a Dexivine that catches you without some weapon to cut it down.
-  Difficulty: [Very Hard Enemies, Fully Random Enemies]
+  Difficulty: [Very Hard Enemies, Fully Random Enemies] # Very Hard due to potential to softlock the player, requiring a restart to escape.
 -
   Pretty name: Keese
   Actor name: keeth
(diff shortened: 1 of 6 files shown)


LagoLunatic:
Typo fixed in c428367764a1f65ed1a6ac4383906b3f3aad6a01

I'm not planning to merge a difficulty setting implementation for enemy rando until it's farther along since things could change.
I also probably wouldn't merge this particular implementation anyway for the reason woofer mentioned.

PR #302, opened 2022-10-30
Disable Map Select in regards to the Spoiler Log

### LINKED ISSUE #301: [Suggestion] Disable Map Select when Spoiler Logs are Disabled
The Map Select tweak has no impact on the Random Seed. As such it is possible in a race Setting to cheat by giving the seed to another player who randomizes the Seed with the Tweak Enabled, who can now go around the map and provide useful information.

Proposed Changes: If "do_not_generate_spoiler_log" is `TRUE`, do not apply the Map Select Tweak. When it is set to `FALSE`, it follows current behavior.

Crain-32:
Resolves #301

### DIFF
--- a/randomizer.py
+++ b/randomizer.py
@@ -381,7 +381,7 @@ class Randomizer:
         tweaks.remove_title_and_ending_videos(self)
       if self.options.get("remove_music"):
         patcher.apply_patch(self, "remove_music")
-      if self.map_select:
+      if self.map_select and not self.options.get("do_not_generate_spoiler_log"):
         patcher.apply_patch(self, "map_select")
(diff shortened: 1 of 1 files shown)


randomizer.py
```diff
@@ -381,7 +381,7 @@ def randomize(self):
         tweaks.remove_title_and_ending_videos(self)
       if self.options.get("remove_music"):
         patcher.apply_patch(self, "remove_music")
-      if self.map_select:
+      if self.map_select and not self.options.get("do_not_generate_spoiler_log"):
         patcher.apply_patch(self, "map_select")
```
wooferzfg, line 385: It might be less confusing if this throws an exception rather than just ignoring the map select patch?

PR #304, opened 2022-12-01
Add three new hint types: Path, Barren, and Location

tanjo3:
This PR introduces three new hint types to the randomizer: path hints, barren hints, and location hints. We draw inspiration for these types of hints from other Zelda randomizers, mainly Ocarina of Time Randomizer. We describe the changes in detail below.

### What's The Same
To begin, we'll go over what hasn't changed. The hint placement options in fishmen, Old Man Ho Ho, and King of Red Lions have not changed, nor has the distribution behavior over these hint placement options. The hints you get from the stone tablet outside Savage Labyrinth, which hint at what is on floors 30 and 50, have not changed. The Wallet mail hint on the items that the Rock Spire Beedle Shop Ship sells has also remained the same. 

Previously, the randomizer used what we will call "item hints", which say that a particular island contains a specific item. These have changed slightly, which we will get into later, but they still exist in addition to the three new hint types. The Two-Eye Octo Fairy hint is an item hint, so it, too, has slightly changed. The user can select the number of each hint type from 0 to 15. The user cannot disable the Savage Labyrinth, Wallet mail, and Octo Fairy hints. So with that, let's introduce the new hint types.

### Path Hints
Path hints are the most complicated hint type, and we'll need to solidify some terminology beforehand.

A **zone** refers to an island, dungeon, the Mailbox region, or the Great Sea region. Each check belongs to a zone. For example, mail checks are in the Mailbox zone. Cyclos is in the Great Sea Zone. Gohma Heart Container is in the Dragon Roost Cavern zone. In this last example, note that the dungeon takes priority over the island. When caves are randomized, the zone of that check is the island where its entrance is. So, if Overlook cave is on Overlook Island, the zone for the check is Overlook Island. However, if Overlook cave is on Pawprint Isle, its zone is Pawprint Isle. Previous item hint implementation only used islands for hinting. So, if something was in Dragon Roost Cavern, the hint would point to Dragon Roost Island (assuming entrances are not randomized). Additionally, hints could not point to Mailbox, the Great Sea, Hyrule, or Ganon's Tower. The new hint types, as well as the item hints, now use the zone name rather than just the island name. There is one special case for zones: if Tower of the Gods leads to a randomized cave, or if something refers to the sunken treasure there, then "Tower of the Gods sector" will be used to differentiate from the dungeon of the same name.

A **progress item** is any item that logically opens up checks in the randomized seed. For example, Triforce shards, swords, and progressive bows are always progress items because reaching Ganondorf and beating him is logically expected in every seed. On the other hand, charts are only progress items if sunken treasure is on. Likewise, Maggie's Letter is only a progress item if Short Sidequests are on. Note that the Maggie's Letter check (Windfall Island - Cafe Bar - Postman) does not have to have a progress item for Maggie's Letter to be considered a progress item. The only condition is that *Windfall Island - Cafe Bar - Postman* is a logical check in that seed's settings.

A **path goal** could be any arbitrary condition in the game, but for path hints, path goals are either:
1. The completion of a race-mode dungeon.
2. The ability to enter Hyrule.
3. The ability to reach and defeat Ganondorf.

For race-mode dungeon path goals, the goal is written as "path to dungeon boss"; for example, a path to Gohma refers to the condition of being able to reach and defeat Gohma for the completion of Dragon Roost Cavern. For the ability to enter Hyrule, the goal is written as "path to Hyrule", and for the ability to reach and defeat Ganondorf, the goal is written as "path to Ganondorf". When race mode is on, only the selected race mode dungeons are valid goals, along with Hyrule and Ganondorf. When race mode is off, only Hyrule and Ganondorf are valid goals.

A **path item** to a path goal is any progress item logically required to achieve that path goal. We write that the item is "on the path to" that path goal. For example, the Grappling Hook is always on the path to Gohma. Likewise, the Triforce shards are always on both the path to Hyrule and the path to Ganondorf. For progressive items, such as bows and wallets, the path item is specific to an instance of that item, not all copies of that progressive item. For example, assume there is only one randomized shard (i.e., you start with seven shards). If that shard is in one of Lenzo's chests, you need one of the progressive Picto Boxes. The progressive Picto Box is a path item to Hyrule if and only if one Picto Box is locked behind the other. So, if there is one Picto Box on Tott and another in the other Lenzo chest, then the Picto Box from Tott is a path item (to Hyrule). However, if one Picto Box is on Tott and the other is on the Salvage Corps, then *neither* Picto Boxes are path items (to Hyrule). This is because you have "a choice" as to which Picto Box you use to reach the shard; therefore, neither is required to reach the path goal. You do need *a* Picto Box, but neither specific instance of a Picto Box is required.

**Path hints** inform the player that a zone contains a path item on the path to a path goal. These path items are selected randomly by randomly picking a valid path goal, and then choosing a path item toward that path goal. When race mode is on, the randomizer tries first to ensure that each race mode dungeon is hinted once and then will select the path goal randomly for the remaining path hints. If an item is on multiple paths, the randomizer will prioritize race mode dungeons, then Hyrule, then Ganondorf. For example, you might get that Windfall Island is on the path to Jalhalla, meaning that a check on Windfall (i.e., on the Windfall zone) gives a path item on the path to Jalhalla. Each path hint always refers to a single item, but that item is unknown to the player. There may be several items on Windfall which are on the path to Jalhalla, which means the player could receive multiple path hints for Windfall to Jalhalla.

Path hints can allow the player to deduce things about the seed logically. For example, suppose you know (by process of elimination, for instance) that a bottle is on the path to Hyrule. In that case, you know that one of the checks locked behind the bottle will lead to a Triforce shard. It could be a direct path, such as a shard being on Grandma mail, or it could be the start of some long chain that eventually results in a shard. As another example, assume you know that the Deku Leaf is on the path to Gohma, that dungeon entrances are not randomized, that Tingle chests are not on, and that key-lunacy is not on. Suppose you find Grappling Hook on some open check, such as Tott. This means that (a) there's a Small Key in the Big Key chest and that (b) Fire and Ice Arrows are locked behind the Deku Leaf. The first deduction is because if there weren't a Small Key in the Big Key chest, you could beat the dungeon logically without the Deku Leaf. The second deduction is that if Fire and Ice Arrows were accessible without the Deku Leaf, it would provide an alternative method to reach the Big Key chest. By definition, a path item is required for that path goal, so if there are alternatives, the Deku Leaf would not be a path item.

Finally, the item on race mode dungeon bosses is never hinted at as path items. Therefore, if you get that Dragon Roost Cavern is on the path to Hyrule, this hint does not refer to the shard (for example) on Gohma Heart Container.

### Barren Hints

**Barren hints** inform the player that the hinted zone contains no path items. We say that the zone is "foolish" or "barren". For example, getting a barren hint for Hyrule means that the Master Sword chamber does not contain a path item. When a race mode dungeon is hinted at as being barren, it excludes the item you get from defeating the dungeon boss. This way, race mode dungeons can still be hinted barren if the rest of the dungeon contains no path items. Furthermore, any mail checks directly unlocked from defeating a barren dungeon are also barren. Non-race mode dungeons are never hinted barren. When entrances are randomized, remember that the zone refers to where the cave's entrance is. So, Overlook Island being hinted barren means that the cave at Overlook Island leads to a barren cave/dungeon. It does not necessarily mean that the Overlook combat cave is barren.

An intuitive way to think about barren hints is that they tell you that you never have to visit that location. Of course, this is a simplified interpretation since you may have to visit the island to complete a race mode dungeon or a sidequest like Goron Trading. The randomizer uses a weighted random selection to pick which barren locations to hint at. Each location is weighted by the square root of the number of checks there. For example, a barren dungeon is typically 3.5-4x more likely to be chosen than a barren location with only one check.

Barren hints can be used to cut down on the number of areas the player is required to visit within a single seed without compromising which settings are on. For example, having eye reefs might be an interesting setting, but doing all six might take a lot of work for every seed. So if you get a few barren hints for some eye reefs, you can still have the experience of doing eye reefs without it becoming too tedious.

### Location Hints

**Location hints** directly tell you about a specific check and what item is there. This item could be a progress item, a path item, or junk. Because of this, the hint will never use cryptic text for the item and will instead hint the item name directly (see [Clearer Hints](#clearer-hints) below). Location hints only hint at certain checks from the list here, and the randomizer selects which checks at random. Only checks which are in logic based on the seed's settings are valid location hints. For example, if Wind Temple is not a race mode dungeon, you will never get a location hint for a check in Wind Temple. The [list of location hints](https://github.com/tanjo3/wwrando/blob/0dc29e454c80fba8639909dfc965e64f9c6a4be4/data/location_hints.txt) was selected to include some of the more annoying and longer checks in the game. This gives each location hint more value than simply selecting checks at random. For example, we don't want to hint at each Free Gift check because that doesn't add anything to the gameplay experience.

Location hints are divided into two categories: sometimes and always. Always hints are a special type of location that will always be hinted, assuming that check is in logic. Always hints are not enabled by default, and can be enabled by checking the "Use Always Hint" setting. For example, Ganon's Maze chest is an always hint, which means that if that check is in logic (i.e., Dungeons is on), the "Use Always Hint" setting in on, and the player specifies to generate at least one location hint, the Ganon's Maze location will always be hinted. The remaining location hints are refered to as sometimes hints, as they will be hinted at sometimes, but not always. We chose the partciularly long and tedious checks to be classified as always hints. The categorization of each location hint can be found in the location hints file, referenced above.

### Item Hints

As mentioned at the start, item hints are the original implementation of hints in the randomizer and have been left largely the same. The only difference is that checks in the Mailbox, the Great Sea, Hyrule, and Ganon's Tower zones can now be hinted at. Note that this affects the Two-Eye Octo Great Fairy hint as well. Previously, only items on islands could be hinted at. Keeping consistent with the previous implementation, if an item is in a dungeon, it will use the island where the entrance is found for the hint location rather than the dungeon itself. The hint text for the new zones are:
* **Mailbox**: the mail
* **The Great Sea**: a location on the open seas
* **Hyrule**: the submerged castle
* **Ganon's Tower**: the stronghold of the evil king

### How Hints Work Together
An important property of the hints is that any particular item can only be hinted at once. So, for example, you can't get two path hints for the same item. So if, for example, you get two path hints for Windfall Island, they must necessarily refer to two different items. In addition, if you get a location hint for a check on Windfall, say that *Windfall Island - Mrs. Marie - Give 40 Joy Pendants* gives the Deku Leaf, you know that neither path hint can refer to that Deku Leaf. The Savage Labyrinth, Wallet mail, and Two-Eye Octo Great Fairy hints are exempt from this. In other words, a path hint for Rock Spire Isle might refer to one of the items that the Rock Spire Beedle Shop Ship is selling.

### Clearer Hints

Based on feedback from the randomizer racing community, we also added the option to enable clearer hints. This replaces the cryptic text in hints with the in-game name for that zone or item. This helps players who have dyslexia or whose native language is not English. Path and barren are always clear, and never use cryptic text. Location hints will always use clearer hints for the item name, regardless if this setting is disabled. Thus, this setting only affects the check name for location hints and the text used in item hints.

### Credits and Final Comments
We like to thank the entire Wind Waker randomizer racing community for their assistance in implementing this PR, both in terms of the technical development of the new hints and in their testing. There were indeed some interesting bugs along the way. The community recently successfully finished the [Season 5 Tournament](https://challonge.com/2kdg2d0o) which featured these new hint types. The Random Settings Tournament's inaugural season is also underway, which uses these hints in an interesting manner. For these races, the settings are randomized and unknown to all racers. Because you can only receive hints for checks which are in logic, the runners can use hints, in combination with Chest Type Matches Contents (CTMC), to logically deduce which settings are enabled to complete the seed efficiently.

Hopefully, that covers everything in this PR. I am open to comments and suggestions on how to improve the feature and how it has been implemented.

### DIFF
--- a/data/location_hints.txt
+++ b/data/location_hints.txt
@@ -82,7 +82,7 @@ Forest Haven - On Tree Branch:
   Text: receiving a gift from the Great Deku Tree
 Forest Haven - Small Island Chest:
   Type: Standard
-  Text: gliding to a remote island near a safe haven
+  Text: gliding to an isle behind the home of the woodland spirits
 
 Forbidden Woods - Mothula Mini-Boss Room:
   Type: Standard
--- a/data/location_hints.txt
+++ b/data/location_hints.txt
@@ -31,25 +31,25 @@ Windfall Island - Maggie's Father - Give 20 Skull Necklaces:
   Text: giving a wealthy man twenty necklaces
 Windfall Island - Zunari - Stock Exotic Flower in Zunari's Shop:
   Type: Sometimes
-  Text: stocking Zunari's shop with three flowers
+  Text: stocking a shop with three flowers
 Windfall Island - Sam - Decorate the Town:
   Type: Sometimes
   Text: volunteering to decorate Windfall
 Windfall Island - Mila - Follow the Thief:
   Type: Sometimes
-  Text: stopping a burglary of Zunari's safe
+  Text: preventing a burglary
 Windfall Island - Battlesquid - Under 20 Shots Prize:
   Type: Always
-  Text: setting a new high score in Battlesquid
+  Text: setting a new record in squid-shooting
 Windfall Island - Pompie and Vera - Secret Meeting Photo:
   Type: Sometimes
   Text: showing a photo to two gossiping ladies
 Windfall Island - Kamo - Full Moon Photo:
   Type: Sometimes
-  Text: snapping a colorful picture of the full moon
+  Text: showing a picture of the moon to a romantic
 Windfall Island - Minenco - Miss Windfall Photo:
   Type: Sometimes
-  Text: showing Minenco a photo of herself
+  Text: showing a lady a photo of herself
 Windfall Island - Linda and Anton:
   Type: Sometimes
   Text: uniting a young couple with the power of photography
@@ -148,19 +148,19 @@ Ganon's Tower - Maze Chest:
 
 Mailbox - Letter from Hoskit's Girlfriend:
   Type: Sometimes
-  Text: the writings of a grateful girlfriend
+  Text: the writing of a grateful girlfriend
 Mailbox - Letter from Baito:
   Type: Sometimes
-  Text: the writings of a bewildered letter sorter
+  Text: the writing of a bewildered letter sorter
 Mailbox - Letter from Orca:
   Type: Sometimes
-  Text: the writings of your hometown swordsman
+  Text: the writing of your hometown swordsman
 Mailbox - Letter from Aryll:
   Type: Sometimes
-  Text: the writings of your beloved sister
+  Text: the writing of your beloved sister
 Mailbox - Letter from Tingle:
   Type: Sometimes
-  Text: the writings of a rupee-hungry fairy
+  Text: the writing of a rupee-hungry fairy
 
 The Great Sea - Cyclos:
   Type: Sometimes
@@ -235,7 +235,7 @@ Tingle Island - Ankle - Reward for All Tingle Statues:
   Text: collecting five glistening fairy figurines
 Tingle Island - Big Octo:
   Type: Sometimes
-  Text: defeating the sea monstrosity at Tingle's Island
+  Text: defeating the sea monstrosity near the island of greed
 
 Bomb Island - Cave:
   Type: Sometimes
--- a/data/location_hints.txt
+++ b/data/location_hints.txt
@@ -258,7 +258,7 @@ Horseshoe Island - Cave:
 
 Flight Control Platform - Bird-Man Contest - First Prize:
   Type: Sometimes
-  Text: gliding to the Flight Control finish line
+  Text: gliding to the finish line of a contest
 Flight Control Platform - Submarine:
   Type: Sometimes
   Text: defeating Wizzrobes in a dimly-lit submarine
--- a/data/location_hints.txt
+++ b/data/location_hints.txt
@@ -96,13 +96,13 @@ Forbidden Woods - Big Key Chest:
 
 Tower of the Gods - Hop Across Floating Boxes:
   Type: Sometimes
-  Text: hopping across boxes in a godly tower
+  Text: hopping across boxes in a tower created to test the hero
 Tower of the Gods - First Chest Guarded by Armos Knights:
   Type: Sometimes
-  Text: defeating two bomb-hungry foes in a godly tower
+  Text: defeating two bomb-hungry foes in a tower created to test the hero
 Tower of the Gods - Darknut Mini-Boss Room:
   Type: Sometimes
-  Text: defeating a sworded foe in its arena in a godly tower
+  Text: defeating a sworded foe in its arena in a tower created to test the hero
 
 Forsaken Fortress - Phantom Ganon:
   Type: Sometimes
--- a/data/location_hints.txt
+++ b/data/location_hints.txt
@@ -106,7 +106,7 @@ Tower of the Gods - Darknut Mini-Boss Room:
 
 Forsaken Fortress - Phantom Ganon:
   Type: Sometimes
-  Text: playing tennis with an apparition in an arena at a forsaken fortress
+  Text: playing tennis with an apparition in the rubble of the evil's lair
 
 Fire Mountain - Cave - Chest:
   Type: Sometimes
--- a/data/location_hints.txt
+++ b/data/location_hints.txt
@@ -1,303 +1,303 @@
 Outset Island - Orca - Give 10 Knight's Crests:
-  Type: Sometimes
+  Type: Standard
   Text: giving a fellow swordsman crests of honor
 Outset Island - Great Fairy:
-  Type: Sometimes
+  Type: Standard
   Text: visiting the fairy atop the place of all beginnings
 Outset Island - Jabun's Cave:
-  Type: Sometimes
+  Type: Standard
   Text: entering the cave at the back of the place of all beginnings
 Outset Island - Dig up Black Soil:
-  Type: Sometimes
+  Type: Standard
   Text: teaching a pig to dig for treasure
 
 Windfall Island - Chu Jelly Juice Shop - Give 15 Green Chu Jelly:
-  Type: Sometimes
+  Type: Standard
   Text: giving a potion master green jelly
 Windfall Island - Chu Jelly Juice Shop - Give 15 Blue Chu Jelly:
-  Type: Always
+  Type: Remote
   Text: giving a potion master blue jelly
 Windfall Island - Mrs. Marie - Give 40 Joy Pendants:
-  Type: Sometimes
+  Type: Standard
   Text: giving a schoolteacher forty pendants
 Windfall Island - Lenzo's House - Become Lenzo's Assistant:
-  Type: Sometimes
+  Type: Standard
   Text: undertaking a photography apprenticeship
 Windfall Island - Lenzo's House - Bring Forest Firefly:
-  Type: Sometimes
+  Type: Standard
   Text: catching a firefly in a bottle
 Windfall Island - Maggie's Father - Give 20 Skull Necklaces:
-  Type: Sometimes
+  Type: Standard
   Text: giving a wealthy man twenty necklaces
 Windfall Island - Zunari - Stock Exotic Flower in Zunari's Shop:
-  Type: Sometimes
+  Type: Standard
   Text: stocking a shop with three flowers
 Windfall Island - Sam - Decorate the Town:
-  Type: Sometimes
+  Type: Standard
   Text: volunteering to decorate a very populated place
 Windfall Island - Mila - Follow the Thief:
-  Type: Sometimes
+  Type: Standard
   Text: preventing a burglary
 Windfall Island - Battlesquid - Under 20 Shots Prize:
-  Type: Always
+  Type: Remote
   Text: setting a new record in squid-shooting
 Windfall Island - Pompie and Vera - Secret Meeting Photo:
-  Type: Sometimes
+  Type: Standard
   Text: showing a photo to two gossiping ladies
 Windfall Island - Kamo - Full Moon Photo:
-  Type: Sometimes
+  Type: Standard
   Text: showing a picture of the moon to a romantic
 Windfall Island - Minenco - Miss Windfall Photo:
-  Type: Sometimes
+  Type: Standard
   Text: showing a lady a photo of herself
 Windfall Island - Linda and Anton:
-  Type: Sometimes
+  Type: Standard
   Text: uniting a young couple with the power of photography
 
 Dragon Roost Island - Rito Aerie - Give Hoskit 20 Golden Feathers:
-  Type: Sometimes
+  Type: Standard
   Text: gathering feathers for a charitable lover
 Dragon Roost Island - Fly Across Platforms Around Island:
-  Type: Sometimes
+  Type: Standard
   Text: gliding around the home of bird people
 Dragon Roost Island - Rito Aerie - Baito:
-  Type: Sometimes
+  Type: Standard
   Text: mastering the art of letter throwing
 Dragon Roost Island - Secret Cave:
-  Type: Sometimes
+  Type: Standard
   Text: besting the cavern at the home of bird people
 
 Dragon Roost Cavern - Mini-Boss:
-  Type: Sometimes
+  Type: Standard
   Text: freeing the chest in an arena at the summit of a large volcanic den
 Dragon Roost Cavern - Under Rope Bridge:
-  Type: Sometimes
+  Type: Standard
   Text: defeating two Bokoblins near a fragile bridge in a large volcanic den
 Dragon Roost Cavern - Big Key Chest:
-  Type: Sometimes
+  Type: Standard
   Text: placing a Magtail on a switch in a large volcanic den
 
 Forest Haven - On Tree Branch:
-  Type: Sometimes
+  Type: Standard
   Text: receiving a gift from the Great Deku Tree
 Forest Haven - Small Island Chest:
-  Type: Sometimes
+  Type: Standard
   Text: gliding to a remote island near a safe haven
 
 Forbidden Woods - Mothula Mini-Boss Room:
-  Type: Sometimes
+  Type: Standard
   Text: exterminating a flying bug in its arena in a dungeon with overgrown vegetation
 Forbidden Woods - Chest in Locked Tree Trunk:
-  Type: Sometimes
+  Type: Standard
   Text: dropping a bomb flower into a tree trunk in a dungeon with overgrown vegetation
 Forbidden Woods - Big Key Chest:
-  Type: Sometimes
+  Type: Standard
   Text: hitting five switches at once in a dungeon with overgrown vegetation
 
 Tower of the Gods - Hop Across Floating Boxes:
-  Type: Sometimes
+  Type: Standard
   Text: hopping across boxes in a tower created to test the hero
 Tower of the Gods - First Chest Guarded by Armos Knights:
-  Type: Sometimes
+  Type: Standard
   Text: defeating two bomb-hungry foes in a tower created to test the hero
 Tower of the Gods - Darknut Mini-Boss Room:
-  Type: Sometimes
+  Type: Standard
   Text: defeating a sworded foe in its arena in a tower created to test the hero
 
 Forsaken Fortress - Phantom Ganon:
-  Type: Sometimes
+  Type: Standard
   Text: playing tennis with an apparition in the rubble of the evil's lair
 
 Fire Mountain - Cave - Chest:
-  Type: Sometimes
+  Type: Standard
   Text: jumping across dried magma to Magtails
 
 Ice Ring Isle - Inner Cave - Chest:
-  Type: Sometimes
+  Type: Standard
   Text: awakening foes from an icy slumber
 
 Earth Temple - Stalfos Mini-Boss Room:
-  Type: Sometimes
+  Type: Standard
   Text: defeating three Stalfos in a haunted dungeon's arena
 Earth Temple - Kill All Floormasters in Foggy Room:
-  Type: Sometimes
+  Type: Standard
   Text: defeating an army of Floormasters in a haunted dungeon
 Earth Temple - Stalfos Crypt Room:
-  Type: Sometimes
+  Type: Standard
   Text: defeating three Stalfos in a crypt deep in a haunted dungeon
 Earth Temple - Big Key Chest:
-  Type: Sometimes
+  Type: Standard
   Text: defeating a sworded foe in a haunted dungeon
 
 Wind Temple - Big Key Chest:
-  Type: Sometimes
+  Type: Standard
   Text: defeating three sworded foes in a a trap-filled dungeon
 Wind Temple - Wizzrobe Mini-Boss Room:
-  Type: Sometimes
+  Type: Standard
   Text: defeating the Wizzrobe mini-boss in a a trap-filled dungeon's arena
 Wind Temple - Chest Behind Seven Armos:
-  Type: Sometimes
+  Type: Standard
   Text: defeating seven mechanical foes near the top of a a trap-filled dungeon
 Wind Temple - Kill All Enemies in Tall Basement Room:
-  Type: Sometimes
+  Type: Standard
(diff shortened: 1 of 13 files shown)


wooferzfg:
Caching the results of `is_requirement_met` is something that was discussed before, and the change wasn't implemented: https://github.com/LagoLunatic/wwrando/pull/177#issuecomment-612646431

However, I think it's more useful in combination with these changes - determining the hints can make seed generation significantly longer, but caching `is_requirement_met` helps a lot with that. @tanjo3 might have the exact numbers on how much faster it is.

(If you would still rather not bother with caching in order to avoid potential bugs, I think everything in this PR should still work without that caching commit.)

tanjo3:
Yeah, so with a sample of 78 seeds with randomized settings, I get:
- with caching: 12.02 ± 4.78 secs.
- without caching: 16.39 ± 6.72 secs.

for hint generation. The majority of this time is in generating the path hints, since the randomizer needs to determine which items are required. As woofer mentions, if it'd be better to remove the caching, that's fine. There are almost certainly other optimizations available to improve hint generation time, which is part of the reason I'd like to get this PR open now.

tanjo3:
Thanks for this. Looking over it, it seems mostly good but I'd suggest the following:

- For path hints, I'd be inclinded to say that "path hints tell you **_that_** an area contains an item that is **_logically_** required to reach a particular **_goal_** in this seed." I'd use "that" here to avoid the implication that all required areas/items will be hinted. I'd use "logically" to specify in what way that item is required. It might be possible to skip the item in the seed by going out of logic, even without glitches (for instance, skipping the Bombs requirement for certain sunken treasures). I'd use "goal" here over "destination" since that's the common terminology for path hints. "Destination" makes sense for goals like accessing Hyrule and Ganon's Tower, but perhaps less so for beating race mode dungeons.
- For barren hints, I'd use "that" over "when" here for the same reasoning as above.
- For always hints, "set high enough" is not quite correct. Even if there's only 1 location hint set to be generated, if there's a location categorized as an always hint in logic, it will be hinted (unless there's more than one such location, in which case, if should be randomly chosen). So there's not really a threshold that has to be reached. Perhaps something like "When **_this setting is selected_**, certain time-consuming locations will always be hinted at, taking precedence over normal location hints" would be better?

LagoLunatic:
>(unless there's more than one such location, in which case, if should be randomly chosen)

Yeah that's what I was trying to get at - it's a bit misleading to say 1 location hint is enough for always locations to "always" be hinted at, since there will usually be more than one location and only one of those will randomly appear. Though my phrasing might have made it even more confusing.

Maybe I should remove the mention of numbers and have the UI grey out that option when the number of location hints is zero instead.

LagoLunatic:
I made path hint generation faster, especially when many locations are selected to be progression (was getting 100+ second generation times, now it's like 10-20 seconds). I don't think it should break anything since I didn't change the actual hint generation logic, but you might want to test it too since you understand path hints better than I do.

hints.py
```diff
@@ -1,166 +1,889 @@
...
+      
+      accessible_locations = self.path_logic.get_accessible_remaining_locations()
+      locations_in_this_sphere = [
+        loc for loc in accessible_locations
+        if loc not in previously_accessible_locations
+      ]
+      if not locations_in_this_sphere:
+        break
+      
+
```
wooferzfg, line 223: There's a few places in this file where there's two blank lines in between lines of code, or in between methods. Not sure if that's intentional.

data/location_hints.txt
```diff
@@ -0,0 +1,303 @@
...
+  Text: catching a firefly in a bottle
+Windfall Island - Maggie's Father - Give 20 Skull Necklaces:
+  Type: Sometimes
+  Text: giving a wealthy man twenty necklaces
+Windfall Island - Zunari - Stock Exotic Flower in Zunari's Shop:
+  Type: Sometimes
+  Text: stocking a shop with three flowers
+Windfall Island - Sam - Decorate the Town:
+  Type: Sometimes
+  Text: volunteering to decorate Windfall
```
wooferzfg, line 37: Should this say "a very populated place" (like the island name hints), or some other generic description, instead of "Windfall"?

hints.py
tanjo3: The code here is based off [`calculate_playthrough_progression_spheres()`](https://github.com/LagoLunatic/wwrando/blob/3d7fc6d61645045bcc9f07dba9204d79df4e04a9/randomizer.py#L878) in `randomizer.py`, so I just used the spacing that's used there.

LagoLunatic:
Hint distribution doesn't work correctly for very low numbers of hints.
For example, on `MS4xMC4wLUJFVEFfMjAyMi0xMS0yOF82NmQyYjlkAEZpbmVUYWNreUJvYXJkAAcBAwA4AIADANAAAgAAAAJAEAAA`, all of the Hohos simply have a blank textbox that says nothing because KoRL already took the only hint.
On `MS4xMC4wLUJFVEFfMjAyMi0xMS0yOF82NmQyYjlkAEZpbmVUYWNreUJvYXJkAAcBAwAsAIADANAAAgAAAAJAEAAA`, hint generation errors out because `update_fishmen_hints` doesn't know how to handle being given zero hints.

When there aren't enough hints I'd probably expect them to be duplicated so that all selected placement options have at least one hint. So if the user wants one hint but all three placement options, that one same hint would be said once by every fishman, once by every Hoho, and once by KoRL.

tanjo3:
I modified the hint distribution behavior to duplicate hints if more placement options are selected than hints generated. It will randomly choose which hint(s) get duplicated so as not to have a preference over which hint type to duplicate (i.e., path > barren > item > location, since that's the order `generate_hints()` returns). It also duplicates each hint once before duplicating a hint again to ensure that it's not the same hint being duplicated over ten placement options despite having generated four hints to begin with, for example. This is just future-proofing since there are only three hint placement options at the moment.


> I made path hint generation faster, especially when many locations are selected to be progression (was getting 100+ second generation times, now it's like 10-20 seconds). I don't think it should break anything since I didn't change the actual hint generation logic, but you might want to test it too since you understand path hints better than I do.

Just wanted to respond to this as well. Thanks so much for doing this. Doing a test on 100 seeds with random settings, I get an average hint generation time improvement from 14.67 ± 4.94 seconds to 3.81 ± 0.56 seconds, which is fantastic! If interested, I'm using the [same weights for settings that we're using for the Random Settings Tournament](https://github.com/tanjo3/wwrando/blob/random-settings/randomizers/random_settings.md). I've also checked through a few seeds to ensure the hint logic is correct, and everything still seems good. 👍

data/location_hints.txt
tanjo3: Resolved with 24211463855546dfc64b594f866448dc907ed48f.

tanjo3:
> When clearer hints are selected and there's a hint for a dungeon item, would it make sense to say something like "Wind Temple small key" (like the [item get text](https://github.com/LagoLunatic/wwrando/blob/master/tweaks.py#L618)) instead of "WT Small Key"?




I could see it. I was initially going to say that using the acronym makes it more readable, but now that I see it, I'm unsure which reads better. The change would make sense from a consistency standpoint. I guess I don't really have an opinion either way.

LagoLunatic:
I think it looks better with the full-length names.

tweaks.py
```diff
@@ -720,6 +727,31 @@ def get_indefinite_article(string):
...
+  elif item_name in DUPLICATABLE_CONSUMABLE_ITEMS:
+    article = get_indefinite_article(item_name)
+  elif item_name.lower().endswith(" small key"):
+    article = get_indefinite_article(item_name)
+  elif item_name in ["Beedle's Chart", "Tingle's Chart", "Maggie's Letter"]:
+    article = None
+  elif item_name.endswith("'s Pearl"):
+    article = None
+  elif item_name in ["Father's Letter", "Moblin's Letter"]:
+    article = get_indefinite_article(item_name)
```
wooferzfg, line 748: Would it make sense to have these be consistent with the textboxes that show up when you get the items?

Screenshot_20221221_004252_Twitch

Screenshot_20221221_004557_YouTube

wooferzfg:
> Path and barren are always clear, and never use cryptic text. Location hints will always use clearer hints for the item name, regardless if this setting is disabled. Thus, this setting only affects the check name for location hints and the text used in item hints.

Just wondering, what was the reasoning for having the "Clearer Hints" setting only affect a subset of the hints, rather than making the hint text either all cryptic or all clear? I personally don't think it's a big deal since the new hint types are in the "Advanced" tab of the settings, but it does seem a bit inconsistent. Also, I think being able to make the Savage Labyrinth and Big Octo hints clearer would be convenient.

tanjo3:
There isn't a real reason behind it other than I didn't find it a priority at the time and never got around to changing it, but I agree that it'd be more consistent with how the other hints work. It would be a minor change; we'd need to add a cryptic text equivalent for the dungeons. Would the path goal for path hints also have cryptic counterparts? So, "path to a gigantic shelled insect" instead of "path to Gohma", for example. On a similar note, should item hints hint at the dungeon then if the item is in a dungeon, or should it just keep hinting at the island where the entrance is. The main reason I kept it that way is to stay consistent with how item hints were previously, but we may as well rework it all. And we may as well include the Savage Labyrinth and Big Octo, as you say.

wooferzfg:
I think that at least making the changes that are relatively easy would be nice to do in this PR, but I wouldn't suggest reworking everything, especially since I assume that most players won't be using these complex hint types. For example, I'm not sure how necessary it is to have cryptic text for path hints, since they're already fairly difficult to understand on their own.

LagoLunatic:
I feel like "Use Always Hints" could be renamed to something that makes it clearer what the option is doing. Maybe "Prefer Out-of-the-way Location Hints"? "Prioritize Long Location Hints"?
And maybe "Clearer Hints" could be "Clarify Hints" or "Simplify Hints" to make it consistent with most other toggles being verbs.
Any opinions on these names?

wooferzfg:
You could also maybe call it "Enable Cryptic Hints" and have it be on by default.

tanjo3:
"Prioritize Long Location Hints" might be ambiguous since it could imply the hints themselves are long, not the locations. However, "Prioritize Long Locations for Location Hints" feels a bit wordy, so I guess I'd just go with "Prioritize Long Location Hints".  Perhaps replace "long" with another adjective, like "tedious" or "inconvenient" to make it clearer that it's the locations that are long?

I also like calling the setting "Enable Cryptic Hints" or "Use Cryptic Text for Hints", and having it on by default.

LagoLunatic:
> You could also maybe call it "Enable Cryptic Hints" and have it be on by default.

That's a good idea.

> "Prioritize Long Locations for Location Hints" feels a bit wordy

"Prioritize Hints for Long Locations"? Though I guess that could imply that it affects other types of hints besides Location Hints.

> Perhaps replace "long" with another adjective, like "tedious" or "inconvenient" to make it clearer that it's the locations that are long?

"Prioritize Tedious Location Hints" was my first thought but that implies monotony which doesn't seem entirely accurate for some of the locations like Shark Island/blue chus.

"Prioritize Inconvenient Location Hints" could work.

What about "Prioritize Remote Location Hints"?

wooferzfg:
"Remote" would probably be my preferred choice, since it means the same as "out of the way", which is probably the best way to summarize the list of prioritized locations overall. "Inconvenient" is fine too, though, in my opinion.

Also, I think "Prioritize Remote Locations for Hints" might be a bit clearer?

tanjo3:
I think "Prioritize Remote Location Hints" would be the best name so that it's clear that this setting only affects location hints. I think that this change should also include modifying [`location_hints.txt`](https://github.com/tanjo3/wwrando/blob/hints-rework/data/location_hints.txt). There, I used "sometimes" and "always" to classify locations, borrowing from Ocarina of Time Randomizer. For consistency, we should probably change this to "location" and "remote". This would mean changing some variable names and such elsewhere in the code as well.

For "Enable Cryptic Hints", thinking about it, might there be a chance a user believes that this setting must be enabled to have hints at all? Wording it as "Use Cryptic Text for Hints" might remove this possible confusion.

LagoLunatic:
When refactoring, I noticed in this part of the code:
```py
  def check_is_legal_item_hint(self, location_name, progress_locations, previously_hinted_locations):
    (...)
    # Remove locations that are included in remote hints.
    if location_name in self.location_hints and self.location_hints[location_name]["Type"] == "Remote":
      return False
```

That remote/always locations get considered as illegal places for an item hint to be, even when the option for prioritizing remote hints isn't enabled (or the number of location hints is too low to guarantee all remote hints get placed).
Was this an oversight? Or an intentional change to make item hints never refer to time-consuming locations? Wasn't sure exactly what the reasoning here was.

Other than that I think this PR looks good and is ready to merge.

tanjo3:
Ah, yeah, that's a bug. That's there because I don't want there to be overlap between item hints and the remote location hints, but the item hints are generated first.

tanjo3:
This commit should address this issue. It was actually a broader issue where I assumed that all remote hints would be generated, even if the user specified to generate less location hints than remote locations available. Now, the randomizer will generate the remote location hints first, and the rest of the hints code will consider that when generating the remaining hints.

LagoLunatic:
Found a bug where some parts of the code were comparing hint text with location names, which could result in duplicate hints because they weren't equal when cryptic text was enabled, so I refactored some more things.

I think everything should be completely functional now, but it might warrant a little bit more testing to be sure nothing's broken.

tanjo3:
One last thing from me: I made a change to the algorithm that determines which zones are barren. The previous algorithm determined junk items and then used special cases to add certain items (Tingle statues, Maggie's Letter, etc.) to that list. This new one determines which items are useful by going backward from the "Can Reach and Defeat Ganondorf" requirement, and develops a list of barren zones from that. This makes it more general, as it no longer needs to consider special cases, so items such as Note to Mom, goddess pearls, etc. can be in barren zones if it's guaranteed that they will only lead to junk.

This list of barren zones is still a superset of "true" barren zones, as all duplicates of progressive items and dungeon keys (in key-lunacy) will never be in a barren location so long as one of those items is useful. This is the case even if one copy of the item locks all the others and the others are not required to beat the seed. However, because this algorithm is more general and the list of barren zones is at least the list of barren zones generated by the old algorithm, this should be better.

Plus, I encounted some bugs with the old algorithm while implementing this one.

hints.py
```diff
@@ -430,198 +430,92 @@ def get_path_hint(self, unhinted_locations, hinted_locations, path_name):
...
+      # Don't consider race mode dungeon bosses, as those are implicity required.
+      if location_name in self.rando.race_mode_required_locations:
+        continue
+      
+      zones_with_useful_locations.add(self.get_entrance_zone(location_name))
+      # For dungeon locations, both the dungeon and its entrance should be considered useful.
+      zone_name, specific_location_name = self.logic.split_location_name_by_zone(location_name)
+      if zone_name in self.logic.DUNGEON_NAMES.values():
+        if location_name != "Tower of the Gods - Sunken Treasure":
+          zones_with_useful_locations.add(zone_name)
```
wooferzfg, line 500: Could you use `self.logic.is_dungeon_location` here?

tanjo3: Yeah, for sure. Thanks for pointing that out. Resolved with 088f0b9e4cddfa0fd9b9a27c7909f2beb064126f.

tanjo3:
Did 100 seeds with random settings again, and there weren't any errors with generating hints besides barren hints (788dfb006863a6dd564a759761d76c20fce73c11). Checked a few of the seeds and everything seemed to be good. For timing, hint generation took, on average, 3.94 ± 0.52 seconds, which is not a significant change since [d6a1d16](https://github.com/LagoLunatic/wwrando/pull/304/commits/d6a1d1680b8795013dea15abbfd1770fb31bee75).

hints.py
LagoLunatic: Is there a reason the similar line in `self.filter_out_hinted_barren_locations` was skipped?
I noticed that one doesn't check the Sunken Treasure special case like the rest.

wooferzfg: Yeah, I think line 570 in `hints.py` should be checking `self.logic.is_dungeon_location` too:
https://github.com/LagoLunatic/wwrando/blob/dbe98824932a2bf8db18478c5e51290cbb76fe76/hints.py#L568-L571

tanjo3: Yeah, I believe you're right. Missed it on the first pass. Resolved with 68d1f8fd35dbb3b438a8fef47e74f6551a9c63ef.

tanjo3:
Thanks a lot to both of you for your assistance and support into getting this PR merged. Happy holidays!

PR #318, opened 2023-01-26
Fixed README for run randomizer command for windows and trailing spaces

ryanakann:


LagoLunatic:
The trailing spaces are an intentional part of GitHub's markdown formatting. Removing them makes the readme look bad.

The version specific in the Python command is unnecessary because wwrando.py includes a shebang with the desired Python version.

ryanakann:
Ack. Cheers

PR #353, opened 2023-11-03
Add Trap Chests to Random Item Pool

Sage-of-Mirrors:
This PR adds the ability for the randomizer to generate trapped treasure chests that play a fun animation without giving the player an item. This is an optional, cosmetic change intended to add variety to randomizer runs.

Integration is 90% complete; the randomizer can generate the chests and adds the custom `event_list.dat` event to maps that require it. The remaining issue to solve is how to distribute the modified treasure chest code; the current patching system is not suited for large-scale changes to RELs.

Trapped chests were made possible by the [TWW decomp project](https://github.com/zeldaret/tww). The modified treasure chest code can be viewed [here](https://github.com/Sage-of-Mirrors/tww/blob/trap_chest_hack/src/d/actor/d_a_tbox.cpp).


randomizers/items.py
```diff
@@ -434,18 +435,27 @@ def change_hardcoded_item_in_rel(self, path, offset, item_name: str):
...
       dzx = self.rando.get_arc(arc_path).get_file("room.dzr", DZx)
     
     chest = dzx.entries_by_type_and_layer(TRES, layer=layer)[chest_index]
-    chest.item_id = item_id
-    
-    if self.options.get("chest_type_matches_contents"):
-      chest.chest_type = self.get_ctmc_chest_type_for_item(item_name)
+
+    if item_name.endswith("Trap Chest"):
+      chest.behavior_type |= 0x40
```
LagoLunatic, line 446: I think there should be a comment mentioning this bit wasn't used by vanilla chests and what it does here

Sage-of-Mirrors: Added!

Sage-of-Mirrors:
Integration is 100% complete - the modified REL has been added to the project as a partially-linked ELF file (PLF), which is converted into a REL at time of randomization.

wooferzfg:
I'm surprised that this setting is off by default, and that there's even a setting for it at all. There's only 5 trap chests placed in 300+ locations, and most of them are likely to end up in non-progress locations anyway.

randomizers/items.py
```diff
@@ -452,9 +452,6 @@ def change_chest_item(self, arc_path: str, chest_index: int, layer: DZxLayer, it
       # Here, we set that custom trap flag.
       chest.behavior_type |= 0x40
 
-      if self.options.get("chest_type_matches_contents"):
-        chest.chest_type = self.rng.randrange(0, 2)
```
wooferzfg: When CTMC is on, will trap chests be the same type as the chest in the vanilla game? That seems a bit strange

Sage-of-Mirrors: Yes, that is the case. I had to remove the randomization here because `self.rng` is invalid at this phase of the process. @LagoLunatic would have to ensure that the object remains valid for me to use it, but they expressed concerns that having access to the RNG object in the saving phase could cause problems.

Sage-of-Mirrors:
@wooferzfg 5 chests was the number that Lago initially suggested, but we could up it to 10 or 15, maybe. As far as the option in the UI, we felt that for now it was best to give the user the option to disable them if they want a "classic" rando experience, lol.

LagoLunatic:
> There's only 5 trap chests placed in 180+ chest locations, and most of them are likely to end up in non-progress locations anyway.

Tried generating 100 seeds (default settings), wound up with 128 traps in progress locations and 330 in non-progress locations. 42 didn't get placed at all since they're currently considered consumables, which aren't guaranteed to be placed.

Traps could be moved to the unique non-progress items list to guarantee they all get placed, but wouldn't help with them usually being in nonprogress locations (the ratio would be basically the same).

I guess they should be considered progress items instead, so they always get placed in progress locations.

> I'm surprised that this setting is off by default, and that there's even a setting for it at all.

Do you think it should be always enabled? It doesn't really seem like something every player would want to me, though maybe it could be enabled by default.

LagoLunatic:
I made sure the traps are always put in progress locations.

The number of traps might need to be increased too, I'm not sure yet. But it should at least feel like more now that they're in places the player will actually check.

wooferzfg:
> Do you think it should be always enabled? It doesn't really seem like something every player would want to me, though maybe it could be enabled by default.

I thought before that it was a really minor change. But since it doesn't work well with Chest Type Matches Contents (the trap chests don't fit the rules that every other item follows) and the trap chests are always in progress locations, it makes more sense for it to be a setting.

LagoLunatic:
> I thought before that trap chests were a really minor change, since there were very few of them and they're very similar to finding a consumable in a chest.

I think the bigger difference than the opportunity cost is the fact that traps damage the player, so you can potentially game over and lose a lot of progress to one if you're on low health and deep in a dungeon. Probably won't happen that often but I think not everyone would like that.

> Chest Type Matches Contents (currently, the trap chests don't fit the rules that every other item follows - I'd suggest making them either always spiky or always wooden)

I missed that the indentation was wrong, should be fixed now so they always appear as progress item chests

PR #354, opened 2023-11-24
Starting item randomizer

Aelire:
This randomizer allows adding up to 3 (completely arbitrary number, probably should change) random progression items to the starting gear of the seed, changing the locations that are accessible from the start of the game

Last year, the racing community experimented with some new options during the season 5 tournament. Some of these features have made it upstream (chest type matches contents, path/barren hints) but one that hasn't is the ability to start the game with additional progression items determined at random.
This is an RFC PR to gauge interest in integrating this in the official build. I'm mostly looking for design feedback and whether you'd be interested in having this feature (not necessarily with this implementation).

* The first commit is preparation to make it slightly easier to manage item groups (eg goddess pearls or tingle statues). This is the smallest refactor that allowed this implementation, but I have another larger version that reworks item groups that I can submit a PR for as well if desired
* The second and third commit are the base implementation
* The fourth commit is a behavior tweak for hints and progression spheres in the spoiler log to avoid the random starting items making other progression items useless for spheres and hints. This can be reverted, in which case the behavior of barren hints changes slightly (it can make items that are completely superseded by the random starting items barren, even if they can also be used to open useful locations)
* The fifth commit is a bugfix to charts handling, where a chart with multiple requirements (eg a triforce chart or a chart in a location with a gunboat, requiring bombs) could be chosen as the last item and make the assertion on new locations fail

<hr>

Testing-wise, i've run bulk generation counting the distribution of random starting items (for 10k seeds) and spot checked a few spoiler logs for those settings, as well as booted up seeds to check the starting inventory is as expected:
* default settings, with 1,2,3 random starting items
* season 5 racing settings with 1,2,3 random starting items
* allsanity with 1,2,3 random starting items
* season 5 settings without dungeons with 2 starting items (to test Farore's Pearl being progression without the other pearls)

For those familiar with the S5 implementation, these are currently the differences compared to that:
* Despite starting with these items, they aren't considered by the logic as starting items to determine which items are progression / CTMC chest types. For a classic example this means that starting with a random boomerang when octos are enabled still keeps a progressive quiver in logic, unlike what the previous build had
* The set of items received can be taken from any progression items with a few exceptions, not just a preset list (this is to make it work with more settings than just the S5/rsl settings)
* Taken all together, the random starting items have to open at least one additional sphere 0 check (for example, no starting with just a magic meter, but also not with just one leaf)

### DIFF
--- a/logic/logic.py
+++ b/logic/logic.py
@@ -189,6 +189,15 @@ class Logic:
         if group_name in self.unplaced_progress_items:
           self.unplaced_progress_items.remove(group_name)
     
+    # Add the randomly-selected starting items (without incidence on other progress items)
+    if self.rando.random_starting_item.is_enabled():
+      for item in self.rando.random_starting_item.random_starting_items:
+        # Needs to happen after make useless_progress_items_nonprogress
+        # To ensure other progress items aren't made nonprogress by the random
+        # items being in the starting inventory for the purpose of hints or
+        # spoiler log progression
+        self.add_owned_item_or_item_group(item)
+
     self.clear_req_caches()
     self.cached_enemies_tested_for_reqs_tuple.clear()
   
--- a/logic/logic.py
+++ b/logic/logic.py
@@ -175,11 +175,7 @@ class Logic:
     self.make_useless_progress_items_nonprogress()
     
     # Replace progress items that are part of a group with the group name instead.
-    for group_name, item_names in self.progress_item_groups.items():
-      if all(item_name in self.unplaced_progress_items for item_name in item_names):
-        self.unplaced_progress_items.append(group_name)
-        for item_name in item_names:
-          self.unplaced_progress_items.remove(item_name)
+    self.unplaced_progress_items = self.pack_item_groups(self.unplaced_progress_items)
     
     # Remove starting items from item groups.
     for group_name, group_item_names in self.progress_item_groups.items():
@@ -251,17 +247,27 @@ class Logic:
     assert location_name in self.item_locations
     self.prerandomization_item_locations[location_name] = item_name
   
-  def get_flattened_unplaced_progression_items(self):
-    progress_items = []
-    for item_name in self.unplaced_progress_items:
+  def expand_item_groups(self, items):
+    ret_items = []
+    for item_name in items:
       if item_name in self.progress_item_groups:
-        group_name = item_name
-        for item_name in self.progress_item_groups[group_name]:
-          progress_items.append(item_name)
+        ret_items += self.progress_item_groups[item_name]
       else:
-        progress_items.append(item_name)
-    
-    return progress_items
+        ret_items.append(item_name)
+
+    return ret_items
+
+  def pack_item_groups(self, items_to_pack: list[str]) -> list[str]:
+    ret_items = items_to_pack.copy()
+    for group in self.progress_item_groups:
+      if all(item_name in items_to_pack for item_name in self.progress_item_groups[group]):
+        ret_items.append(group)
+        for item_name in self.progress_item_groups[group]:
+          ret_items.remove(item_name)
+    return ret_items
+
+  def get_flattened_unplaced_progression_items(self):
+    return self.expand_item_groups(self.unplaced_progress_items)
   
   def get_num_progression_locations(self):
     return Logic.get_num_progression_locations_static(self.item_locations, self.options)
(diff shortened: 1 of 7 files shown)


randomizers/starting_item.py
```diff
@@ -0,0 +1,147 @@
...
+    if self.options.get("sword_mode") == "No Starting Sword" and "Progressive Sword" in available_items:
+      available_items.remove("Progressive Sword")
+
+    # without shuffling here, the chosen items sometimes stay stable for the
+    # same seed with different starting_gear options as a prefix of the list may
+    # be the same.
+    # We need to sort before converting to lists because sets have per-process
+    # hash seeding and could otherwise lead to different randomization between
+    # processes with the same seed
+    return self.rng.sample(sorted(available_items), k=len(available_items))
```
LagoLunatic, line 133: Is this call the rng.sample necessary? You can make a set deterministic just by calling sorted, not sure what is accomplished by sampling the entire list and then randomly choosing from it, and the comment doesn't explain it either.

Aelire: It's not necessary for correctness, I can drop it if undesirable

Without it, the only call(s) to the rng is the `self.rng.choice` to select the actual item, and the list that gets passed to it is fairly stable (there isn't a huge amount of variation in progression items, at least with the settings I've tried), so with the same seed but changing settings, you would still very often get the exact same set of starting items. The `sample` makes it so adding or removing a single item from the `available_items` list results in completely different selected items even with the same seed

I don't know how much you care about things changing with the same seed when changing settings, so this is possibly just not relevant and can be dropped

LagoLunatic: I actually usually try to aim for the opposite of what you described, so that changing settings while keeping the seed the same changes the randomization as little as possible. For example, I explicitly reset the RNG seed in between each separate randomizer option being run to minimize how much they affect each other and make it mostly depend on the seed (which is probably part of the reason you saw these starting items being the same with different settings). Though the options affect each other in so many indirect ways that it's not really possible to make it depend entirely on the seed in most cases, so this isn't an explicit feature.

Also, for cases where making a list more random actually is desired, I would just call rng.shuffle on that list and then pop one from it, which is probably more readable and performant than sampling the entire list from itself and then choosing one from it.

Aelire: Understood! I'll drop it and update the PR

logic/logic.py
```diff
@@ -454,7 +483,8 @@ def get_items_by_usefulness_fraction(self, item_names_to_check):
     accessible_undone_locations = self.get_accessible_remaining_locations(for_progression=True)
     inaccessible_undone_item_locations = []
     locations_to_check = self.remaining_item_locations
-    locations_to_check = self.filter_locations_for_progression(locations_to_check)
+    need_sunken_treasure = any(item.startswith("Treasure Chart ") or item.startswith("Triforce Chart ") for item in item_names_to_check)
+    locations_to_check = self.filter_locations_for_progression(locations_to_check, filter_sunken_treasure=(not need_sunken_treasure))
```
Aelire, line 487: This function is only called here outside of the starting item randomizer: https://github.com/LagoLunatic/wwrando/blob/7375ab52e4b3fd9bcf180081fa930ce5f98dcaf5/randomizers/items.py#L282-L287

I'm not completely sure of all the implications of this change but I haven't seen anything obvious in testing so far. I think the only times this would make a difference to that callsite in the items randomizer is:
* If no charts are in logic, then chart locations aren't considered for progression, so the grappling hook doesn't always get a max fraction of 2 regardless of settings (I think this is a bugfix)
* If some charts are in logic but all charts have already been placed, and the grappling hook hasn't yet, then it artificially has a lower usefulness fraction (but since it is required for ganondorf it's still <9999)

A maybe simpler/safer solution (that I'm not able to write/test properly right now, it might break other things) could be to instead in `filter_locations_for_progression_static` actually follow the treasure/triforce charts progression options and always exclude the sunken locations when the locations aren't enabled, in addition to filtering them with filter_sunken_treasure for the callers that need it

LagoLunatic: I think the safest solution would be just adding a new `filter_sunken_treasure` parameter to `get_items_by_usefulness_fraction` that it passes down to `filter_locations_for_progression`. Then the item randomizer can pass False guaranteeing that nothing changes about the existing behavior there for now, and the starting items randomizer can pass whatever it needs to.

I'm curious about this apparently being a bugfix for the grappling hook in the item randomizer but that should be a separate PR later, I don't know what all the implications here are either since I haven't thought about this part of the code in 5 years, and I wasn't able to find a seed that seems affected by the bug you describe. But the item randomizer tends to be pretty fragile in edge cases.

Aelire: > I'm curious about this apparently being a bugfix for the grappling hook in the item randomizer

I don't think it really can realistically affect the item randomizer (and I also haven't managed to find a seed that exposes it, it would be extremely rare), I just meant a bug in that the fraction for the grappling hook returned by `get_items_by_usefulness_fraction` might take into account charts that aren't actually in logic and be artificially low

I'll test a bit more, see if I can wrangle a seed that reproduces it and send the PR if I find one

Aelire:
Thanks for merging this! Feel free to ping me for bug reports if any come up

PR #358, opened 2023-12-19
Pack item groups when checking enemy requirements

Aelire:
Adding/removing progress group items individually breaks the invariant that `Logic.unplaced_progress_items` always contains full groups.

In the original refactor of the logic for item groups I didn't test the enemy randomizer. 
When handling item combos the enemy randomizer breaks down groups into their individual components to add/remove them, which makes `unplaced_progress_items` contain all the individual items from the group instead of the group name.
Later when the item randomizer adds/removes all items to place dungeon items, it tries to remove the group name, instead of each individual item. Because the group name isn't in `unplaced_progress_items` (instead all the individual items are), `remove_owned_item_or_item_group` takes the branch that removes the whole group and tries to remove each individual item from `currently_owned_items` even though they aren't in the list, which raises.

Example permalink that this fixes: (reported in discord): `MS4xMC4wX2Q1MTRhMjIAR2F5RW5lbXlSYW5kbwDHPEDR0PxqVf0BAAAAAAAAAAAAAA==`

Fixes: 4f0f52b22661ddde9bb76e31f1fc366524a9a734

### DIFF
--- a/logic/logic.py
+++ b/logic/logic.py
@@ -1325,7 +1325,7 @@ class Logic:
     biggest_combo = []
     for item_name, num in max_num_of_each_item_to_check.items():
       biggest_combo += [item_name]*num
-    biggest_combo = tuple(biggest_combo)
+    biggest_combo = tuple(self.pack_item_groups(biggest_combo))
     # print(f"Biggest item combo length: {len(biggest_combo)} items")
     # print(f"Biggest item combo: {biggest_combo}")
     


LagoLunatic:
This is changing enemy rando logic to add/remove only entire groups at once? Wouldn't that technically be inaccurate logic for combos where the player has found just some of the items in the combo? It's supposed to check every possible combination of items the player might actually own at this point, not combinations of groups the item randomizer might place.

I guess it won't actually affect anything since Farore's Pearl is currently the only group item that is useful without the rest of its group, but ideally this code should continue to function even if new groups are added in the future, so the individual items in a group should be independent.

Why did the extra starting items rando require this refactor in the first place, anyway?

Natolumin:
I think, but I'd need to double-check, that since you kept commit 7c0b9b489c3440b0fe0b7da958ea06d56a2d74e3 as well, there wouldn't be an issue with reverting this refactor.
I don't completely remember off the top of my head but I believe the issue this handled is that anything in `WWRandomizer.starting_items` is removed from `Logic.progress_item_groups` during initialization, but the extra starting item randomizer modified `starting_items` after the logic is initialized

Aelire:
I think, but I'd need to double-check, that since you kept commit 7c0b9b489c3440b0fe0b7da958ea06d56a2d74e3 as well, there wouldn't be an issue with reverting this refactor.
I don't completely remember off the top of my head but I believe the issue this handled is that anything in `WWRandomizer.starting_items` is removed from `Logic.progress_item_groups` during initialization, but the extra starting item randomizer modified `starting_items` after the logic is initialized

Aelire:
> I guess it won't actually affect anything since Farore's Pearl is currently the only group item that is useful without the rest of its group, but ideally this code should continue to function even if new groups are added in the future, so the individual items in a group should be independent.

I don't think this was actually true before the refactor in question either; when reverting it, during the temporary items addition/deletion (https://github.com/LagoLunatic/wwrando/blob/d514a22afdd54d577b718af51a3bc7d71823fc55/logic/logic.py#L1403C1-L1403C1), the `unplaced_progress_items` variable contains both the group *and* the individual items:
```
['Delivery Bag', 'Ballad of Gales', "Earth God's Lyric", 'Song of Passing', 'Note to Mom', "Maggie's Letter", "Moblin's Letter", 'Ghost Ship Chart', 'Progressive Wallet', 'Progressive Picto Box', 'Progressive Picto Box', 'Empty Bottle', 'DRC Big Key', 'DRC Small Key', 'DRC Small Key', 'DRC Small Key', 'DRC Small Key', 'FW Big Key', 'FW Small Key', 'TotG Big Key', 'TotG Small Key', 'TotG Small Key', 'ET Big Key', 'ET Small Key', 'ET Small Key', 'ET Small Key', 'WT Big Key', 'Triforce Shards', 'Goddess Pearls', 'Tingle Statues', 'Progressive Shield', 'Progressive Shield', 'Power Bracelets', "Wind God's Aria", 'WT Small Key', 'WT Small Key', 'Boomerang', 'Progressive Bow', 'Progressive Bow', 'Progressive Bow', 'Progressive Sword', 'Bombs', 'Progressive Sword', 'Progressive Magic Meter', 'Progressive Sword', 'Triforce Shard 1', 'Triforce Shard 2', 'Triforce Shard 3', 'Triforce Shard 4', 'Triforce Shard 5', 'Triforce Shard 6', 'Triforce Shard 7', 'Triforce Shard 8', 'Command Melody', 'Iron Boots', 'Deku Leaf', 'Progressive Magic Meter', 'Progressive Sword', 'Skull Hammer', 'Hookshot', 'Grappling Hook']
```

This doesn't seem to be causing immediate issues, but my impression when writing the initial `pack/unpack_item_groups` code was that these variables always needed to have the whole group, but not all code properly enforced that

LagoLunatic:
The groups probably need to be refactored completely, they only need to exist in the item randomizer because of how hard it is to get forward fill to generate a seed under hyper-restrictive conditions, but I built groups into the logic class when item rando was the only feature that existed without thinking about it too much. I'll see if I can remove them from everything other than the item randomizer.

Aelire:
I think that'd be great, I certainly hit my head against it a couple times. I did try that refactor initially when writing the extra_starting_item randomizer, and got to somewhere that roughly works at least on "normal" settings but I haven't tested it very extensively, so putting it here in case that can help: https://github.com/Aelire/wwrando/commit/d29c9127aaf37cb5fe45d1ec2b3000578b5ebe81

Aelire:
> I think, but I'd need to double-check, that since you kept commit [7c0b9b4](https://github.com/LagoLunatic/wwrando/commit/7c0b9b489c3440b0fe0b7da958ea06d56a2d74e3) as well, there wouldn't be an issue with reverting this refactor.

I finally double-checked, and the main issue the refactor (in 4f0f52b) handled was that `Logic.add_owned_item_or_item_group` didn't update the `Logic.unplaced_progress_items` variable at all. 
If the `extra_starting_items` randomizer chooses a group, then said group gets assigned twice, once as an extra starting item, and then the group also gets placed in progression locations. Example spoiler log (on master + reverted 4f0f52b):

```
Wind Waker Randomizer Version 1.10.0_f6053b2
Permalink: MS4xMC4wX2Y2MDUzYjIAMQBJEFAYAIB9AXQJAAAAAAQAggAADA==
Seed Hash: Sam Missy
Seed: 1
Options selected:
  progression_dungeons, progression_puzzle_secret_caves, progression_great_fairies, progression_free_gifts, progression_expensive_purchases, progression_misc, sword_mode: Start with Hero's Sword, num_required_bosses: 4, logic_obscurity: None, logic_precision: None, mix_entrances: Separate Dungeons From Caves & Fountains, hoho_hints, fishmen_hints, num_item_hints: 15, num_location_hints: 5, cryptic_hints, swift_sail, instant_text_boxes, reveal_full_sea_chart, skip_rematch_bosses, remove_title_and_ending_videos, starting_gear: ['Ballad of Gales', 'Progressive Magic Meter', 'Progressive Shield', 'Song of Passing'], num_extra_starting_items: 3, custom_player_model: Link, custom_color_preset: Default


Random starting item: Tingle Statues, Progressive Wallet, Command Melody

Playthrough progression spheres:
1:
  Outset Island:
      Mesa the Grasscutter's House:                Goddess Tingle Statue
  Windfall Island:
      Jail - Tingle - Second Gift:                 Boomerang
      Maggie - Free Item:                          Grappling Hook
      Tott - Teach Rhythm:                         Din's Pearl
  Dragon Roost Island:
      Wind Shrine:                                 Forbidden Tingle Statue
  Dragon Roost Cavern:
```

So I don't think you can cleanly revert that commit, or it would need the extra_start_items randomizer to do its own maintenance of the `unplaced_progress_items` list. The item randomizer doesn't run into that because it calls into `set_multiple_locations_to_group`, which does maintain the `unplaced_progress_items` list itself here: https://github.com/LagoLunatic/wwrando/blob/d514a22afdd54d577b718af51a3bc7d71823fc55/logic/logic.py#L249C52-L249C52

LagoLunatic:
I removed the groups completely: b610ce489132e05b7388bbbf0fd7692fd6961b77

Not only were they way buggier and harder to refactor properly than I originally assumed, but it doesn't seem like they're actually accomplishing their original purpose anymore either, so it's better to just remove them. This might cause some other bugs to appear so let me know if anything's weird.

PR #369, opened 2024-04-05
Very minor changes

Crain-32:
Started working on the next major version of multiworld, started a larger PR of stuff before cutting it back to some very simple stuff confirm on things before continuing.

Things I'm looking at doing to the main branch, but want confirmation before doing them.
Currently only `wwrando_options.Options` extends `BaseOptions`. I'm thinking of collapsing `BaseOptions` into `Options` to help consolidate some context.

In doing that I'd also swap from the current `all` from `BaseOptions` to instead be something more like
The Entrances/Exits with
```python
  def __repr__(self):
    return f"ZoneExit('{self.unique_name}')"
  
  all: ClassVar[dict[str, 'ZoneExit']] = {}
  
  def __post_init__(self):
    ZoneExit.all[self.unique_name] = self
```

Swap `if/elif` statements with `match/case` statements.
```python
# Connect signals to detect when the user changes each option.
if isinstance(widget, QAbstractButton):
  widget.clicked.connect(self.update_settings)
elif isinstance(widget, QComboBox):
  widget.currentIndexChanged.connect(self.update_settings)
  if option.choice_descriptions:
    widget.highlighted.connect(self.update_choice_highlighted_description)
elif isinstance(widget, QListView):
  pass
elif isinstance(widget, QSpinBox):
  widget.valueChanged.connect(self.update_settings)
else:
  raise Exception("Option widget is invalid: %s" % option.name)
# Instead
match widget:
  case QAbstractButton(): widget.clicked.connect(self.update_settings)
  case QComboBox(): 
    widget.currentIndexChanged.connect(self.update_settings)
    if option.choice_descriptions:
   # Rest of the direct cases/logic
   case _:
       raise Exception("Option widget is invalid: %s" % option.name)
```

Last one would be the lack of  `__init__.py` in a lot of folders that are treated more like modules. Feels a bit weird, so I was wondering the rational behind that.

### DIFF
--- a/options/base_options.py
+++ b/options/base_options.py
@@ -1,6 +1,7 @@
-from dataclasses import fields, MISSING, Field, _recursive_repr, _FIELDS, _FIELD, asdict
+from dataclasses import fields, MISSING, Field, _FIELDS, _FIELD, asdict
 from typing import Any, Type, Optional
 import typing
+from reprlib import recursive_repr
 
 class Option(Field):
   type: Type
@@ -25,7 +26,7 @@ class Option(Field):
     self.hidden = hidden
     self.unbeatable = unbeatable
   
-  @_recursive_repr
+  @recursive_repr
   def __repr__(self):
       return ('Option('
               f'name={self.name!r},'
--- a/options/wwrando_options.py
+++ b/options/wwrando_options.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass, KW_ONLY
+from dataclasses import dataclass
 from enum import StrEnum
 
 from options.base_options import BaseOptions, option
--- a/randomizers/extra_starting_items.py
+++ b/randomizers/extra_starting_items.py
@@ -1,11 +1,10 @@
-import typing
 from logic.item_types import DUNGEON_PROGRESS_ITEMS
 
 from randomizers.base_randomizer import BaseRandomizer
 import tweaks
 from options.wwrando_options import SwordMode
 
-DISALLOWED_RANDOM_STARTING_ITEMS = set([
+DISALLOWED_RANDOM_STARTING_ITEMS = {
   # There's a separate option for number of starting triforce shards that we'd need to mess with.
   "Triforce Shard 1",
   "Triforce Shard 2",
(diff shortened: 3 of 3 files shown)


LagoLunatic:
> Things I'm looking at doing to the main branch, but want confirmation before doing them.
> Currently only `wwrando_options.Options` extends `BaseOptions`. I'm thinking of collapsing `BaseOptions` into `Options` to help consolidate some context.

Why? Is there some benefit to this? I intentionally split them into two files so that the metaprogramming stuff is in a different file from the list of options for organization.

> In doing that I'd also swap from the current `all` from `BaseOptions` to instead be something more like
> The Entrances/Exits with

BaseOptions.all and ZoneExit.all are not the same thing.
ZoneExit.all is a list of every *instance* of a ZoneExit (which is a dataclass, so it has a `__post_init__` method).
BaseOptions.all is a list of every *field* of a BaseOptions. Fields are not dataclasses, they have no `__post_init__` method.
So your proposed code isn't functional.

> Swap `if/elif` statements with `match/case` statements.

Maybe, I can try it out and see if it works the same and looks better.

> Last one would be the lack of `__init__.py` in a lot of folders that are treated more like modules. Feels a bit weird, so I was wondering the rational behind that.

Why do you want `__init__.py`? Is something not working for you unless you add it? If so, please file a bug report explaining the problem.

options/base_options.py
```diff
@@ -1,6 +1,7 @@
-from dataclasses import fields, MISSING, Field, _recursive_repr, _FIELDS, _FIELD, asdict
+from dataclasses import fields, MISSING, Field, _FIELDS, _FIELD, asdict
 from typing import Any, Type, Optional
 import typing
+from reprlib import recursive_repr
```
LagoLunatic, line 4: What's the reasoning behind adding another import when recursive_repr already exists in dataclasses?

Crain-32: https://github.com/python/cpython/blob/558b517bb571999ef36123635e0245d083a28305/Lib/dataclasses.py#L248 Since dataclasses is just copying the implementation. Importing it from original location makes it easier to identify where to find documentation on it, since you have to go all the way to the Python Source to know where to find the actual documentation.

LagoLunatic: But I don't want documentation on the function, I just want it to behave the same way as it does in dataclasses. The dataclasses implementation and the reprlib implementation seem slightly different, and the point is just to extend dataclasses fields with a subclass, so I copied everything as closely as possible.

Crain-32:
> Why? Is there some benefit to this? I intentionally split them into two files so that the metaprogramming stuff is in a different file from the list of options for organization.

> BaseOptions.all is a list of every field of a BaseOptions. Fields are not dataclasses, they have no __post_init__ method.
So your proposed code isn't functional.

Goals are to remove indirection or remove code we can have Python handle for us since we've gotten to 3.12, which has a lot of improvements.

Maybe [this](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields) is worth looking into? Since `Options(BaseOptions)` is a dataclass, just not `Option(Field)`. `BaseOptions.by_name` is a bit trickier if you don't want to add any functions to the `Options(BaseOptions)` class. But this could replace the `BaseOptions.all` property.

> Maybe, I can try it out and see if it works the same and looks better.

Yeah it's mainly a cosmetic change, the main issue I've found is that if you do something like
```python
case MyClass() as myInstance:
```
Then at least PyCharm views `myInstance` as type `any`.

> Why do you want __init__.py? Is something not working for you unless you add it? If so, please file a bug report explaining the problem.

I'm not biased one way or the other, I've just not seen it before. I've heard that "best practice" is the `__init__.py` stuff, but best practice is

options/base_options.py
Crain-32: It also seems `@recursive_repr`'s implementation didn't work with `Field` as well.

LagoLunatic:
> Maybe [this](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields) is worth looking into?

That's what `BaseOptions.all` already uses. Using it directly would work too, it would just require an extra import.

> Then at least PyCharm views myInstance as type any.

I don't use pycharm, but the vscode extension (pylance/pyright) seems to correctly detect the type of the instance regardless of whether `isinstance` or `match/case` are used.

> I'm not biased one way or the other, I've just not seen it before. I've heard that "best practice" is the `__init__.py` stuff, but best practice is

I had heard `__init__.py` was necessary for module imports in some cases and I used to have them. But I removed them a few months ago because it turned out they're unnecessary and iirc they were making pyinstaller builds more difficult to get working than necessary. So I think this "best practice" isn't actually correct.

PR #374, opened 2024-04-16
Remove low health animation at 1 heart max health

Aelire:
Follow-up to #370, discord user noah_and_toad pointed me to the addresses for `checkRestHPAnime` as a way to disable the "tired" animation link has at <=1.5 hearts, which I couldn't figure out how to do in the original PR.
This adds a shim and check to that function to behave similarly to the low health beep, so that it only triggers if link has lost 0.5 hearts from max health

I tried to better follow the existing style of asm patches so you have less fixups to do when merging, do you use any automated formatting or rules for these that I could use to make it easier for you?


LagoLunatic:
> I tried to better follow the existing style of asm patches so you have less fixups to do when merging, do you use any automated formatting or rules for these that I could use to make it easier for you?

I'm not sure specifically what you're referring to since I don't remember changing much of the style of the last asm patch you PRed. Most of the changes I made were practical, like removing all the nops and implementing a new .area directive for the assembler in order to prevent any future minor changes to those asm patches causing issues like crashes or corruption as a result of how it was originally written meaning that adding a single new line would cause overwriting code.

I don't like autoformatters and don't really have any style rules, except I guess "I think hexadecimal is more readable than decimal".

PR #389, opened 2024-06-06
Allow user to exclude locations

tanjo3:
This PR adds an "Excluded Locations" tab that lets users specify progression locations to be ignored by the randomizer.

The interface here is similar to selecting starting items from randomized items. On the "Excluded Locations" tab, the user can view all the progression locations for the settings selected from the main tab. They can use the buttons to move locations to the "Excluded Locations" side; locations on this side will be considered non-progress locations by the randomizer. Additionally, two filters are added for both progression and excluded locations, allowing the user to quickly find a location in the list. The filters do not affect the final value for either location list.

A few things to note:

- If you have randomized charts on, all 49 sunken treasure locations will be considered progression locations, since we have yet to determine which chart maps where. If you select sunken treasure locations to be excluded, they will contain non-progress items, and the charts leading to those locations will also be considered non-progress.
- Building on that, items solely required to unlock excluded locations will be considered non-progress. For example, even if Miscellaneous is on, if the Ankle check is excluded, Tingle statues will be regarded as non-progress. This means they will be in wooden chests for the purposes of chest type matches contents.
- Excluded locations are considered nonprogress and are not included in hint generation. For example, excluding Ganon's Maze chest means it will never appear as a location hint.
- If you have dungeons on but exclude too many checks in a dungeon so the randomizer can't place dungeon items, you will get an error.
- This is not meant to be an interface for selecting which bosses are required for Required Bosses Mode.

I've also added zlib compression, alongside base64 encoding, for permalinks. This is optional. However, since each location adds a bit to the bitstring, it can get rather long. For the default settings and a seed name of A:
Before: `MS4xMC4wXzBjMjYyNmQAQQBJEFAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGAAA+wLIJQAAAAAAAQBBAIAA`
After: `eJwz1DM00DOIN0g2MjMyS2FwZPAUCGAgEkgwMPxmOqEKZjMCtTYwAAA+Dgbz`

If any further revisions or clarifications are needed, please let me know.


tanjo3:
I've added 10 tests that hopefully cover a sufficient number of setting combos to test this feature.

test/test_dry.py
```diff
@@ -41,6 +42,8 @@ def enable_all_progression_location_options(options: Options):
 
 def enable_all_options(options: Options):
   enable_all_progression_location_options(options)
+  # options.progression_locations
+  # options.excluded_locations
```
wooferzfg, line 46: Did you mean to leave these comments in?

tanjo3: Yeah, I figured it was in the style of the test function to have all the options there, but the unmodified ones commented out. Like with `randomized_gear`/`starting_gear`.

wooferzfg:
I got an error when running the [Windows build](https://github.com/LagoLunatic/wwrando/actions/runs/9439187875/artifacts/1583442849):

Wind_Waker_Randomizer_4CZGSuPJp1

LagoLunatic:
I fixed the pkg error on master, you can revert the workaround in wwrando.spec

PR #390, opened 2024-06-14
Fix nonrandomized dungeon entrances with randomized boss entrances and required bosses mode

Aelire:
Currently, the combination of randomizing boss entrances but not dungeon entrances when using required bosses mode mostly fails to randomize successfully.
Example permalink: `MS4xMC4wXzM5NTRjZTkAQQAFEEAUAAH7AugSAAAgAIAAhCAAQAA=`.
A bulk test with these settings seems to have a 100% failure rate, but other values for number of required bosses (except 6) also fail to randomize most of the time

The first commit resolves the spurious `Not enough island entrances left to split entrances` in this case
The second commit resolves an underlying bug now made reachable where nonrandomized dungeons weren't tracked in `islands_with_a_banned_dungeon`, occasionally tripping the `not banned_island_names & required_island_names` assertion (same permalink, about 10% failure rate)

### DIFF
--- a/randomizers/entrances.py
+++ b/randomizers/entrances.py
@@ -486,6 +486,18 @@ class EntranceRandomizer(BaseRandomizer):
     
     self.randomize_one_set_of_exits(remaining_entrances, remaining_exits, terminal_exits)
   
+  def check_if_one_exit_is_progress(self, exit: ZoneExit) -> bool:
+    locs_for_exit = self.zone_exit_to_logically_dependent_item_locations[exit]
+    assert locs_for_exit, f"Could not find any item locations corresponding to zone exit: {exit.unique_name}"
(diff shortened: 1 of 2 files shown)


wooferzfg:
Probably worth adding the settings you mentioned to the [tests](https://github.com/LagoLunatic/wwrando/blob/master/test/test_dry.py)?

Aelire:
Removed an unrelated change from the first commit. 
It addressed a randomization failure when randomizing dungeon entrances but not boss entrances, ie the opposite situation, where a required dungeon and banned dungeon could still end up on the same island with 2 entrances, but I initially forgot these 2 issues were different.
I'll send a separate PR for that second issue

LagoLunatic:
This also seems to fix the same bug that happens when randomizing inner secret cave entrances but not outer secret cave entrances. Thanks!

PR #391, opened 2024-06-14
Fix a randomization failure with nested entrances and required bosses mode

Aelire:
When placing a banned entrance nested inside a secret cave inner isle or a dungeon, the top-level island wouldn't be marked as containing a banned dungeon, potentially letting a required boss/dungeon be placed on the same island later.
Recurse through the entrances in that case to ensure the banned island list is maintained correctly.

This fixes an assertion error(`assert not banned_island_names & required_island_names`) with example permalink `MS4xMC4wXzZhMzc4ZjgAMjE0AEGfUQRgC/sC6BIAAAAACAAEAQAC`

### DIFF
--- a/randomizers/entrances.py
+++ b/randomizers/entrances.py
@@ -700,12 +700,19 @@ class EntranceRandomizer(BaseRandomizer):
       self.done_entrances_to_exits[zone_entrance] = zone_exit
       self.done_exits_to_entrances[zone_exit] = zone_entrance
       
-      if zone_exit in self.banned_exits and zone_entrance.island_name is not None:
+      if zone_exit in self.banned_exits:
         # Keep track of which islands have a required bosses mode banned dungeon to avoid marker overlap.
         if zone_exit in DUNGEON_EXITS + BOSS_EXITS:
           # We only keep track of dungeon exits and boss exits, not miniboss exits.
           # Banned miniboss exits can share an island with required dungeons/bosses.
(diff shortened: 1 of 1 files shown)


wooferzfg:
Probably worth adding a [test](https://github.com/LagoLunatic/wwrando/blob/master/test/test_dry.py) for this too

Aelire:
This one only has a ~0.15% failure rate with these settings, so a test with a fixed seed would become useless on any commit that changes randomization. 
The same settings also have other rarer failure modes that this doesn't attempt to fix (eg seed 3462/`MS4xMC4wXzZhMzc4ZjgAMzQ2MgBBn1EEYAv7AugSAAAAAAgABAEAAg==` fails differently: `No valid exits to place for entrance: Dungeon Entrance on Dragon Roost Island`) which could cause false positives.

Alternatively you'd need to run about 4000 seeds in the test to get down to a 1% false-negative rate (which imo is still too high for a statistical test, while probably taking way too much time)

Aelire:
Added a second case (fairly rare too) triggering the same assertion, where randomizing entire dungeons but not nested entrances could lead to placing required and unrequired dungeons on the same quadrant in required bosses mode (because the code only checked for banned boss entrances rather than boss + dungeon)
Example permalink: `MS4xMC4wXzkzMmFiNzEAMTE0AA0AQBRgCPgC4BIAAAAAgACAIABAAA==`. This has an about ~2% failure rate in bulk testing

PR #392, opened 2024-06-14
RFC: Add an option to restore the bomb flower puzzle blocking DRC entrance

Aelire:
Looking for some feedback on whether this would be interesting / considered for inclusion. There are some remaining issues with dungeon-only start explained below that I'm trying to figure out good solutions for, but if this is a no-go in the first place I can spare some work :)

Currently the randomizer removes the puzzle at the entrance to DRC where you have to blow up a boulder to fill up a pond with water to cross a gap. This means that the DRC entrance is always accessible from the very start of the game with no items.
Comments on the patch imply that this was made because the boulder is counter-intuitively only breakable with bomb flowers, rather than the usual bombs or power bracelets. 
Replacing this special boulder (Eskban) with a normal one (Ebrock) makes the puzzle much more intuitive, so a few people in the racing community have considered restoring this puzzle with this change to add a bit of variety

### Randomisation failures

* It's possible with this option to craft a seed with enough progression locations in logic but no location accessible from the game start, even without entrance randomizer (eg dungeon-only seed with "Open DRC" unchecked and default starting items). 
This was already possible without this option with dungeonless seeds (eg `MS4xMC4wXzZhMzc4ZjgAQQAAIx8AAAj4AuASAAAAAAgABAEAAg==`), so I'm not sure if this is considered a regression

* With entrance randomizer, the "safety entrance" mechanism doesn't take into account starting items, so seeds that could potentially randomize successfully by having enough starting items to open one of the dungeon entrances would fail to randomize
I'm planning to send a couple additional changes (still wip) to the entrance randomizer to successfully choose a safety entrance more dynamically, reducing (hopefully eliminating) the spurious randomization failures that currently happen with this change (but I can't do anything for the seeds where it's legitimately impossible)

### Din's Pearl

In 1.6.0, the effect of Din's pearl on the lava pool in front of the entrance was removed, so that there's always a bomb puzzle to solve to enter DRC. Since this PR adds another puzzle anyway, I removed this patch when the boulder is present (additionally, it makes it easier to cross with the leaf, as it calms the wind, making Din's pearl a useful QoL item)
I'm happy to remove that if not desired


wwr_ui/randomizer_window.ui
```diff
@@ -1171,6 +1171,31 @@
               </layout>
              </widget>
             </item>
+            <item>
+             <widget class="QGroupBox" name="groupBox_logic_tweaks">
+              <property name="title">
+               <string>Logic Tweaks</string>
```
wooferzfg, line 1177: Maybe this should be called something different since it affects things in game and not just the logic? This option seems similar to "Skip Boss Rematches", but isn't really a convenience tweak

wooferzfg:
With this setting being in the Advanced Options tab, hopefully the additional randomization failures are okay. It seems like a tricky problem to avoid randomization failures in every case

PR #394, opened 2024-06-18
Add "Hint Importance" option

tanjo3:
This PR adds a new hint option called "Hint Importance". When this setting is enabled, item and location hints are slightly modified to include information if the reward is required, possibly required, or not required. This setting has been added somewhat recently to other randomizers (Majora's Mask and Twilight Princess randomizers, for instance). I'm using the naming convention from the OoTMM cross randomizer. This is an optional setting, disabled by default.

If this setting is disabled, then hints are written as before. If this setting is enabled, generated item and location hints may be modified with this importance information. This information is already determined when generating path and barren hints, so no additional logic processing needs to be done.

An item that is **required** is an item that is required for at least one of the path goals in the randomized seed. An item that is **not required** is an item that the player does not need to collect to complete all the path goals. An item that is **possibly required** is an item that may be used to accomplish a path goal, but there may be a different path to that goal that does not use that item. This follows from the path and barren logic of hints already in the randomizer.

Below are two examples of item hints (on the King of Red Lions), showing examples of items being hinted as possibly required and not required:
  

In the second example, the hinted Tingle statue is not required because the item on the Ankle check (which is the only check directly locked behind Tingle statues) is also not required. However, the Ankle check is a logical location in the seed (i.e., the Miscellaneous category of locations is on). Non-progress items do not include importance information, even with the hint importance options turned on, because they are trivially non-required. A quick benefit of the setting is that if sunken treasure from Triforce Charts and cryptic hints are _off_, the reader can quickly tell the difference between a hinted Triforce **_Shard_** and Triforce _**Chart**_ since a hint for a shard will note that it is required and a hint for a chart will contain no additional information. This is a frequent misread for runners. Additionally, note the color highlighting of the hint importance text.

Here is an example of a location hint for a required item:

Since the Savage Labyrinth and Octo Fairy hints are item and location hints, respectively, they, too, are affected by the setting. The Savage Labyrinth stone tablet follows a similar syntax for other location hints:

The colors for the tablet text are different, so I couldn't quite get the same palette, but the highlighting is the important part (for ease of readability), not necessarily the exact colors.

The Octo Fairy's text is modified to incorporate the importance, as shown below:

The original text (i.e., with the hint importance option off) would say, "A piece of the power of the gods, which may help you on your quest." If hint importance is on and the hinted item is not required, the following text box would say, "Though if you still desire such an item, you must journey to that place." Otherwise, the original text of "When you find you have need of such an item, you must journey to that place." is used. I made this change since saying that an item is not required and then saying "when you need the item..." sounds awkward.

Some additional implementation notes: since the Savage Labyrinth and Octo Fairy require the hint importance information, and this information is calculated in the call to `generate_hints()` in `randomizers/hints.py,` I've moved this call to before either the Savage Labyrinth and Octo Fairy hints are generated. This doesn't affect hint generation. Also, dungeon keys are conventionally not considered path items when key-lunacy is disabled since getting a path hint for the DRC Small Key in the first room would feel pretty bad (as an example). However, if a location containing a dungeon key is selected for a location hint, we don't want to say that the key is not required when it obviously is. As such, the filter for dungeon keys in non-key-lunacy seeds is moved to after `get_required_locations_for_paths()` and all path locations (including keys) have been recorded in `self.path_locations`. Finally, since remote location hints are necessarily determined before path or barren hints, the setting of the importance of these items is deferred until both path and barren hints are generated.

### DIFF
--- a/randomizers/hints.py
+++ b/randomizers/hints.py
@@ -986,7 +986,7 @@ class HintsRandomizer(BaseRandomizer):
     
     return remote_hintable_locations, standard_hintable_locations
   
-  def get_location_hint(self, hintable_locations):
+  def get_location_hint(self, hintable_locations, ignore_importance=False):
     if len(hintable_locations) == 0:
       return None
     
@@ -995,7 +995,10 @@ class HintsRandomizer(BaseRandomizer):
     hintable_locations.remove(location_name)
     
     item_name = self.logic.done_item_locations[location_name]
-    item_importance = self.get_importance_for_location(location_name)
+    if ignore_importance:
+      item_importance = None
+    else:
+      item_importance = self.get_importance_for_location(location_name)
     
     location_hint = Hint(HintType.LOCATION, location_name, item_name, item_importance)
     
@@ -1053,7 +1056,8 @@ class HintsRandomizer(BaseRandomizer):
     if self.prioritize_remote_hints:
       remote_hintable_locations, standard_hintable_locations = self.get_legal_location_hints(progress_locations, [], [])
       while len(remote_hintable_locations) > 0 and len(hinted_remote_locations) < self.max_location_hints:
-        location_hint, location_name = self.get_location_hint(remote_hintable_locations)
+        # Temporarily ignore item importance until later.
+        location_hint, location_name = self.get_location_hint(remote_hintable_locations, ignore_importance=True)
         
         hinted_remote_locations.append(location_hint)
         previously_hinted_locations.append(location_name)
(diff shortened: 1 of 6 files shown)


randomizers/hints.py
```diff
@@ -12,22 +12,29 @@
...
+  BARREN = 2
+  ITEM = 4
+  LOCATION = 8
+  FIXED_LOCATION = 16
+
+
+class ItemImportance(Flag):
+  REQUIRED = 1
+  POSSIBLY_REQUIRED = 2
+  NOT_REQUIRED = 4
```
LagoLunatic, line 26: Why is Flag being used instead of Enum for these two? You don't seem to be combining multiple together. And you also don't seem to be using bitwise operators or `in`.

tanjo3: Oh, good catch. I had used the `Flag` property while working on the feature, but didn't use it in the end. I can revert those changes.

PR #408, opened 2025-02-18
Fix bug in path hint calculation

tanjo3:
This PR fixes a bug in path hint calculation.

To determine if an item is path, the randomizer plays through the seed, removing that item from the playthrough, and then determining if the path goal is still reachable. If it is, then that item must not be required, and therefore not path. Otherwise, it must be a path item. However, during this playthrough, the randomizer will also consider locations in non-required dungeons when required bosses mode is on. So, if a duplicate copy of an item is accessible there, then the copy that is in logic may no longer be considered a path item, despite it being required for the playthrough. This only arises for items like Empty Bottles whose copies may be placed in non-progress locations.

Also, the filter for progress locations used to be done at the end of each sphere. We've moved it up in the loop. This way, the randomizer doesn't need to determine the accessibility of non-progress locations for each sphere, as these should not be considered for path logic in the first place.

### DIFF
--- a/randomizers/hints.py
+++ b/randomizers/hints.py
@@ -665,7 +665,10 @@ class HintsRandomizer(BaseRandomizer):
     while self.path_logic.unplaced_progress_items:
       progress_items_in_this_sphere = {}
       
-      accessible_locations = self.path_logic.get_accessible_remaining_locations(for_progression=False)
+      # Only consider progress locations.
+      accessible_locations = self.path_logic.get_accessible_remaining_locations(for_progression=True)
+      accessible_locations = list(set(accessible_locations) - set(self.rando.boss_reqs.banned_locations))
+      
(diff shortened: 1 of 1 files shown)


LagoLunatic:
Do you have an example seed where this occurs? Mostly I'm curious if this affects the progression spheres shown in the spoiler log as well, which has similar logic to this.

LagoLunatic:
Tried on master and yeah it seems pretty common with the right settings, around 10% of the time, e.g. MS4xMC4wXzc3ODA5MGMAMTk3AAk4EAwAAPsCyCUAAAAAAAEAQQCAAA==
And even when it doesn't happen the nonrequired dungeons appearing at all in the spheres is kinda confusing. I'll have to remove them from that section entirely.

PR #417, opened 2025-12-31
Add more shuffle options for dungeon items

tanjo3:
This PR adds more shuffle options for dungeon items.

Previously, the options were to either have _all_ dungeon items shuffled within their own dungeons or to shuffle _all_ dungeon items anywhere (key-lunacy). This PR first splits the dungeon items into three categories, each with its own shuffle option: small keys, Big Keys, and Dungeon Maps and Compasses.

The new shuffle options are:
- **Vanilla**: All items in that category will be in their vanilla (original) locations.
- **Start With**: The player starts with all items in that category in their inventory.
- **Own Dungeon**: All items in that category will be shuffled within progression locations in their original dungeon. This is the original default behavior of the randomizer.
- **Any Dungeon**: All items in that category will be shuffled within any progression location in any dungeon. If Required Bosses Mode is on, dungeon items of required bosses' dungeons will not appear in non-required bosses' dungeons.
- **Overworld**: All items in that category will be shuffled throughout progression overworld locations.
- **Anywhere**: All items in that category will be shuffled in any progression location. This is the original default behavior of key-lunacy. 

The UI of the randomizer has been updated to provide three dropdowns to allow the user to select these shuffle modes _independently_ of one another. The default is having all dungeon items shuffled to their own dungeons. The hint logic of the randomizer has also been updated to only hint at keys if they are shuffled in any dungeon, overworld, or anywhere. Similarly, only small keys shuffled in vanilla or own dungeon modes will not count as their own sphere.

The randomized starting gear list has also been updated to include dungeon keys, along with the maps and compasses. I also changed the list's behavior to filter out dungeon items when their shuffle mode was set to vanilla or start-with shuffle mode. Starting with the dungeon items is self-explanatory and is really a visual bugfix. I chose to remove the items in vanilla shuffle mode to avoid needing additional logic for which vanilla locations get randomized. For instance, with vanilla small keys and a DRC Small Key added to the starting gear, which DRC Small Key vanilla location gets randomized? It's not immediately obvious which is the best way to handle this, as there are multiple approaches.

Finally, I've added some tests for combinations of dungeon item shuffle modes. The first test method covers all combinations with default settings. The rest cover combinations of key shuffle modes in Required Bosses Mode with 4 (default) and 1 boss, and when dungeons are not progression. I did some more testing locally, but this is already 324 new tests, which should cover most cases.
**Edit:** Since tests were taking over 2 hours to run, I cut down the number of tests to 158 by removing maps and compasses as a parameter. I also maked most tests added in this PR as "thorough" so that they can still be run locally, but GH Actions will only run 36 of the newly-added tests.


wooferzfg:
I'm curious what the reasoning was for bosses, and minibosses when those entrances are randomized, to not have dungeon items. These were the commits that made these changes:
- https://github.com/LagoLunatic/wwrando/commit/efbec974e49a5e0fa05e1953dfcbbd238dcf9d22
- https://github.com/LagoLunatic/wwrando/commit/6271b412bf6f632bca5095a727b98ea447c921f0

tanjo3:
I updated the randomized starting gear list to include dungeon keys, along with the maps and compasses. I also changed the list's behavior to filter out dungeon items when their shuffle mode was set to Vanilla or Start With. Start With is self-explanatory and is really a visual bugfix. I chose to remove the items in Vanilla shuffle mode to avoid needing additional logic for which vanilla locations get randomized. For instance, with vanilla small keys and a DRC Small Key added to the starting gear, which DRC Small Key vanilla location gets randomized? It's not immediately obvious which is the best way to handle this, as there are multiple approaches.

